PR #6, opened 2024-04-18
feat: add lwr packages / fix commit hooks

nrkruk:
### What does this PR do?
Add lwr dependencies to the cli repo

[@W-15428922@](https://gus.lightning.force.com/a07EE00001nnGsvYAE)

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,18 +6,16 @@
   "bugs": "https://github.com/forcedotcom/cli/issues",
   "dependencies": {
     "@oclif/core": "^3.26.4",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -136,8 +138,7 @@
         "test:command-reference",
         "test:deprecation-policy",
         "lint",
-        "test:json-schema",
-        "link-check"
```
nrkruk, line 140: link-check seems to want a published CLI command which we don't have yet obviously

nrkruk: Removing for now

PR #42, opened 2024-05-30
feat: initial revision for app preview on desktop

maliroteh-sf:
### What does this PR do?

Adds support for `sf lightning preview app` command for desktop only.

To run the command, ensure that you have authorized at least one org:
- `sf lightning preview app` => should launch the default org and navigate to the default app
- `sf lightning preview app -n Sales` => should launch the default org and navigate to the Sales app
- `sf lightning preview app -o MyOrg` => should launch the specified org and navigate to the default app
- `sf lightning preview app -o MyOrg -n Sales ` => should launch the specified org and navigate to the Sales app

@W-15429062@


command-snapshot.json
```diff
@@ -1,15 +1,15 @@
 [
```
maliroteh-sf, line 1: auto-generated after running `yarn update-snapshots`

package.json
```diff
@@ -6,22 +6,23 @@
   "bugs": "https://github.com/forcedotcom/cli/issues",
   "dependencies": {
```
maliroteh-sf, line 7: Since I was pulling in `@salesforce/lwc-dev-mobile-core` I also decided to update the rest of the dependencies to the latest version.

schemas/lightning-preview-app.json
```diff
@@ -1,16 +0,0 @@
-{
```
maliroteh-sf, line 1: I removed this file since for now `LightningPreviewApp` extends from `SfCommand<void>` and does not return anything. As we work on adding more functionalities to `LightningPreviewApp`, we can change the return type as needed and auto-generate the new schema

src/commands/lightning/preview/app.ts
```diff
@@ -1,41 +1,137 @@
...
 
-export default class LightningPreviewApp extends SfCommand<LightningPreviewAppResult> {
+export default class LightningPreviewApp extends SfCommand<void> {
   public static readonly summary = messages.getMessage('summary');
   public static readonly description = messages.getMessage('description');
   public static readonly examples = messages.getMessages('examples');
 
   public static readonly flags = {
     name: Flags.string({
       summary: messages.getMessage('flags.name.summary'),
```
maliroteh-sf, line 36: As per latest discussion [here](https://salesforce.quip.com/lw7LAqDyuhjn#TUKADArZyof), `--name` is now optional.

src/commands/lightning/preview/app.ts
```diff
@@ -1,41 +1,137 @@
...
+    }
+  }
+
+  private async desktopPreview(logger: Logger, flags: LightningPreviewAppFlags): Promise<void> {
+    const appName = flags['name'];
+    const targetOrg = flags['target-org'];
+    const platform = flags['device-type'];
+
+    logger.debug('Determining Local Dev Server url');
+    // todo: figure out how to make the port dynamic instead of hard-coded value here
```
maliroteh-sf, line 73: As the work for automatically starting the dev server progresses, we should have a better idea on how to achieve this.

messages/lightning.preview.app.md
```diff
@@ -4,7 +4,7 @@ Preview Lightning Experience Applications.
 
 # description
 
-Preview components, org, and sites. If no topic is specified, the default action is to preview the org.
+Preview components, apps, and sites. If no topic is specified, the default action is to preview the org.
```
khawkins, line 7: A couple of things on:

> If no topic is specified, the default action is to preview the org.

- Is there precedent for this? Having default behavior at the topic hierarchy level? I personally find that confusing, based on the fact that the different subtopics have different input parameters. We should confirm.
- If that behavior is confirmed, we need to change the default action verbiage to something like "...to preview the most recently accessed Lightning App".

src/shared/orgUtils.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import { Connection } from '@salesforce/core';
+
+export class OrgUtils {
+  /**
+   * Given an app name, it queries the org to find the matching app id. To do so,
+   * it will first attempt at finding the app with a matching DeveloperName. If
+   * no match is found, it will then attempt at finding the app with a matching
+   * Label. If multiple matches are found, then the first match is returned.
+   *
```
maliroteh-sf, line 16: See [this discussion](https://salesforce.quip.com/lw7LAqDyuhjn#TUKADAFeHPe) which lead to the implementation here.

messages/lightning.preview.app.md
```diff
@@ -21,11 +21,23 @@ Use the appropriate topic to preview specific aspects of the development environ
...
 # flags.name.summary
 
-Description of a flag.
+Name of the Lightning Experience application to preview.
 
-# flags.name.description
+# flags.target-org.summary
 
-More information about a flag. Don't repeat the summary.
+Specify the org to preview.
```
khawkins, line 28: We should probably use some variation of [the verbiage](https://github.com/search?q=org%3Asalesforcecli%20%22Username%20or%20alias%20of%20the%20target%20org%22&type=code) that other CLI plugins use, for this description.

sfdctaka (approved):
LGTM!

messages/lightning.preview.app.md
khawkins: Oh wait, I'm realizing that this is verbiage *specifically* for `sf lightning preview app`. It's the *beginning of the description* that threw me off. I'll make an edit suggestion.

khawkins: Ah I see that these values have already been defined in the spec. This one just has the wrong verbiage. Should be:

```suggestion
Preview Lightning Experience Applications in real-time.
```

src/commands/lightning/preview/app.ts
khawkins: Cc: @nrkruk

src/commands/lightning/preview/app.ts
```diff
@@ -1,41 +1,137 @@
...
+    const ldpServerUrl = PreviewUtils.generateWebSocketUrlForLocalDevServer(platform, '8081');
```
khawkins, line 74: @nrkruk et al: I just want to call out that with this line we are (in this moment anyway) officially putting LDP-related utility functionality into the [lwc-dev-mobile-core](https://github.com/forcedotcom/lwc-dev-mobile-core) project/package. I think such functionality could live in there, or we could create a new independent package, or we could re-brand the existing one. No action necessary here: I just wanted to call it out for discussion.

khawkins, line 74: Independently: I see that we're passing down `platform` as a `string`. Should `lwc-dev-mobile-core` instead own the `Platform` enum, and have the accompanying method signature?

src/commands/lightning/preview/app.ts
```diff
@@ -1,41 +1,137 @@
...
+      const appId = await OrgUtils.getAppId(targetOrg.getConnection(undefined), appName);
+      if (!appId) {
+        throw new Error(messages.getMessage('error.fetching.app-id', [appName]));
+      }
+
+      logger.debug(`App Id is ${appId}`);
+
+      appPath = `lightning/app/${appId}`;
+    } else {
+      logger.debug('No Lightning Experience application name provided.... using the default app instead.');
```
khawkins, line 97: Nit: this is probably a good candidate for `logger.info()` instead. Wouldn't hurt to routinely notify the user of their choice here.

src/commands/lightning/preview/app.ts
```diff
@@ -1,41 +1,137 @@
...
+    // org to be forwarded to OrgOpenCommand.
+    //
+    // Or we could simply look at the raw arguments passed to the LightningPreviewApp command,
+    // find the raw value for --target-org flag and forward that raw value to OrgOpenCommand.
+    // The OrgOpenCommand will then parse the raw value automatically. If the value is
+    // valid then OrgOpenCommand will consume it and continue. And if the value is invalid then
+    // OrgOpenCommand simply throws an error which will get bubbled up to LightningPreviewApp.
+    //
+    // Here we've chosen the second approach
+    const idx = this.argv.findIndex((item) => item.toLowerCase() === '-o' || item.toLowerCase() === '--target-org');
```
khawkins, line 118: Does it require a lot of parsing of the `Org` data? Or does an `Org` get fleshed out with all of its appropriate configuration data, once its parameter input value has been resolved to an object?

If it's the latter, then we could simply use `org.getUsername()` on the resolved org, and use that as the value to be passed to `org open`.

maliroteh-sf: `org.getUsername()` alone is not enough since it returns `string | undefined` and in case it returns undefined then we should look into pulling out the org alias name. The alias is available on `org.options` but the problem is that `options` is private. I could work around it by using `(org as any).options.aliasOrUsername` but there are 2 issues:

1. It's hacky and can break in the future if the fields of `options` change
2. `options.aliasOrUsername` is also nullable

Ideally we would do a contribution to `@salesforce/core` and add a method to `Org` that would have the logic of resolving the org identity for us. But for now I resorted to the simplest approach. There is plenty more work to be done in this repo and on the `preview app` command so we can always revisit and improve this if needed.

src/commands/lightning/preview/app.ts
maliroteh-sf: Sure... but just FYI `logger.blah` will only output to `~/.sf/log-file` and not to `stdout`. I didn't want to have anything extra printed to `stdout` b/c I was trying to follow Marcelino's doc with the example output.

src/commands/lightning/preview/app.ts
maliroteh-sf: Good idea, I think we should move it to `lwc-dev-mobile-core`... there's plenty more work to be done on the `preview app` command in future PRs so once I move this to `lwc-dev-mobile-core` and we publish a new version of it, I will update this code in future PRs.

src/shared/orgUtils.ts
```diff
@@ -0,0 +1,41 @@
...
+   * @param connection the connection to the org
+   * @param appName the name of the app
+   * @returns the app id or undefined if no match is found
+   */
+  public static async getAppId(connection: Connection, appName: string): Promise<string | undefined> {
+    // NOTE: We have to break up the query and run against different columns separately instead
+    // of using OR statement, otherwise we'll get the error 'Disjunctions not supported'
+    const devNameQuery = `SELECT DurableId FROM AppDefinition WHERE DeveloperName LIKE '${appName}'`;
```
khawkins, line 24: Is `LIKE` overly permissive for these queries? Should we not use `=`?

maliroteh-sf: No it is not. `LIKE` allows us to check while ignoring cases. It also can be used for pattern matching `but only if` we add pattern wildcards such as `%` Since here I am not adding any wildcards then `LIKE` will strictly match the name but ignoring the casing.

khawkins (commented):
This all looks pretty good to me. Just a couple of questions.

src/commands/lightning/preview/app.ts
khawkins: Oh okay. If it's all going to a log file anyway, it's probably less important.

src/commands/lightning/preview/app.ts
khawkins: Oh yeah, if you basically _don't_ get a fully configured Org data object by virtue of passing in the seeding value, then I agree: let's just go with the approach you have. My hope was that all got fleshed out for you behind the scenes, but sounds like no.

khawkins (approved):
Looks good!

src/commands/lightning/preview/app.ts
khawkins: One way or another we should be deferring the org parsing logic to the CLI functions we're calling. That shouldn't be this plugin's process to implement.

PR #49, opened 2024-06-04
feat(dev-server): integrate with preview app command

rax-it:
### What does this PR do?
Add flow to start `lwc-dev-server` when running `sf lightning preview app`

@W-15044710@


src/lwc-dev-server/index.ts
```diff
@@ -0,0 +1,76 @@
...
+      return LogLevel.error;
+    case 60:
+      return LogLevel.silent;
+    default:
+      return LogLevel.error;
+  }
+}
+
+function createLWCServerConfig(source: string, logger: Logger): ServerConfig {
+  const rootDir = path.resolve(source, 'force-app/main/default');
```
sfdctaka, line 42: Shouldn't the config be placed elsewhere? My guess is this path reflects a Salesforce VSCode project, as shown in the attached screenshot. The config is for LWR server and LWC server, right? Then, would the root folder be a more appropriate place for the config file to be created?

image

maliroteh-sf: I think the config file will be at the root folder. The value for `rootDir` usually indicates to the server as to where the components can be found.

And looking at line 58 I see that `process.cwd()` is being used as the base path (which I think would be the same as the root directory for an SFDX project)

README.md
```diff
@@ -183,6 +183,4 @@ FLAG DESCRIPTIONS
     This person can be anyone in the world!
 ```
 
-_See code: [src/commands/hello/world.ts](https://github.com/salesforcecli/plugin-lightning-dev/blob/1.0.25/src/commands/hello/world.ts)_
```
rax-it, line 186: Lint was complaining about broken links, README looks out of date

src/lwc-dev-server/index.ts
rax-it: `lwc-dev-server` supports multiple workspaces and the config is used to figure out where does the components live, sfdx project [workspace=sfcli in this case] follows an opinionated folder structure and isn't required to have `lwr.config` or either `lwc.config`. But let me know if my understanding is incorrect

sfdctaka: The way I'm reading this code is that it is creating a `ServerConfig` object in `<root dir>/force-app/main/default`. I assume the object that's created inside this method is eventually going to be replaced by values that are read from `lwc.config.json` in that folder.

You say in your comment that `lwc.config(.json)` isn't required in the folder. Yes, the code is `resolve`-ing it, implying that there exists a json there.

Here's what I'm trying to do for myself, @rax-it . I'm working on this repo too and I'm trying to figure out where the `lwc.config.json` exists. As for where it should exist, that, I don't know. So when I found your draft PR and read the code I naturally thought `<root dir>/force-app/main/default` is the folder and trying to confirm with you.

maliroteh-sf: @sfdctaka I think I may have understood the root of your confusion here. There are 2 concepts of root directory, each serving a different purpose.

For an SFDX project, the root directory of the entire SFDX project is the folder where the file `sfdx-project.json` is located at. `lwr.config.json` should also be placed at this same location.

Now inside the `lwr.config.json` there are various configuration flags/values. One of those flags is `rootDir` and it means a different thing. `rootDir` tells the dev server where to look for when resolving modules/components.

Suppose you have an SFDX project located at `~/Desktop/MyProject` and that this project has an lwc component called helloWorld (which would be located at `~/Desktop/MyProject/force-app/main/default/lwc/helloWorld`.

By setting `rootDir = ~/Desktop/MyProject/force-app/main/default` you are telling the dev server that, when requests come in to serve helloWorld (and any other components/modules that it may be referencing) then the dev server should start looking at the location defined by `rootDir` to search for the components.

If you have further questions on this, feel free to reach out to me on slack (so we don't pollute the comments/feed of this PR)

@rax-it The assumption that for SFDX projects, the modules/components will always be under `force-app/main/default` is not true. We would need to look at the content of `sfdx-project.json` and process `packageDirectories` where it would tell us where the packages are. I've seen code for processing `sfdx-project.json` a long while ago and can't remember off the top of my head what it would involve. I'll look around to see if I can find it and will post link(s) here if I do.

rax-it: @maliroteh-sf I added sfdx config handling as per what information was available in [developer docs](https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_ws_mpd.htm).
Let me know if you find more info about it.

maliroteh-sf: Yes that sounds about right. I also was looking at [sfdxPackageDirsPattern](https://github.com/forcedotcom/lightning-language-server/blob/6363a7b4040db51c249de097efb7cbf01e289256/packages/lwc-language-server/src/base-indexer.ts#L16) which is subsequently used for example by [getModulesDirs](https://github.com/forcedotcom/lightning-language-server/blob/6363a7b4040db51c249de097efb7cbf01e289256/packages/lightning-lsp-common/src/context.ts#L314) or by [componentEntries](https://github.com/forcedotcom/lightning-language-server/blob/6363a7b4040db51c249de097efb7cbf01e289256/packages/lwc-language-server/src/component-indexer.ts#L51)

src/lwc-dev-server/index.ts
```diff
@@ -0,0 +1,101 @@
...
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { existsSync, lstatSync, readFileSync } from 'node:fs';
+import path from 'node:path';
+import process from 'node:process';
+import { LWCServer, LogLevel, ServerConfig, Workspace, startLwcDevServer } from '@lwc/lwc-dev-server';
+import { Logger } from '@salesforce/core';
+
+const DEV_SERVER_PORT = 8081;
```
ravijayaramappa, line 14: The port needs to be configurable. This a pretty popular port number and we should anticipate that it might now be available. Can we add a new flag to `sf lightning preview app` command?

src/lwc-dev-server/index.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  return {
+    rootDir,
+    port: DEV_SERVER_PORT,
+    protocol: 'wss',
+    host: 'localhost',
+    paths: namespacePaths,
+    workspace: Workspace.SfCli,
+    targets: ['LEX'], // should this be something else?
```
ravijayaramappa, line 75: This field is not used anywhere actually. I think this value is safe to keep for now as-is.

src/lwc-dev-server/index.ts
maliroteh-sf: @ravijayaramappa we may need to run it by Marcelino to discuss this. For example:

- Should it be required or optional flag
- If it is optional and omitted what should we default too
- What if the user uses a `--port` flag to pass in a port but `lwc.config.json` also has a port defined in it. Should we update the content of the config file and overwrite with the new flag?
- etc.

Maybe we should utilize our slack channel to discuss further with him and the devs?

src/lwc-dev-server/index.ts
```diff
@@ -0,0 +1,101 @@
...
+        logger.warn(`Skipping ${resolvedDir} because it does not exist or is not a directory`);
+      }
+    }
+  }
+
+  return {
+    rootDir,
+    port: DEV_SERVER_PORT,
+    protocol: 'wss',
+    host: 'localhost',
```
ravijayaramappa, line 72: @rax-it The host and protocol must be remove from the ServerConfig definition in lwc-platform-public. They are un-necessary to be set. You can proceed with this PR but please create a WI to remove them in lwc-platform-public

src/lwc-dev-server/index.ts
```diff
@@ -0,0 +1,101 @@
...
+      lwcDevServer = null;
+    }
+  };
+
+  // normal exit flow
+  process.on('exit', cleanup);
+  // when a user presses ctrl+c
+  process.on('SIGINT', cleanup);
+  // when a user kills the process
+  process.on('SIGTERM', cleanup);
```
ravijayaramappa, line 98:
```suggestion 
  [
     'exit', // normal exit flow
     'SIGINT',  // when a user presses ctrl+c
     'SIGTERM',  // when a user kills the process
   ].forEach(signal => process.on(signal, cleanup));
```

test/lwc-dev-server/index.test.ts
```diff
@@ -0,0 +1,46 @@
...
+// eslint-disable-next-line no-underscore-dangle
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const logger = {
+  debug: () => {},
+  warn: () => {},
+  trace: () => {},
+  getLevel: () => 10,
+} as Logger;
+
+describe('lwc-dev-server', () => {
```
ravijayaramappa, line 24: Can we have a test which can use a sample sfdx project and start the lwc dev server and stop it successfully? Like a sanity end-2-end test.

ravijayaramappa (approved):
LGTM, just a few minor suggestions and testcase request

src/lwc-dev-server/index.ts
ravijayaramappa: I like that idea of just putting it in the `lwc.config.json`. So that the user can add it once and forget it. Created W-15962937 to track.

src/lwc-dev-server/index.ts
rax-it: W-15963048

maliroteh-sf (approved):

PR #50, opened 2024-06-06
feat: add support for previewing on mobile

maliroteh-sf:
### What does this PR do?

Adds support for `sf lightning preview app` command for mobile.

To run the command, ensure that you have authorized at least one org:
- `sf lightning preview app -o MyOrg -n MyApp -t ios -i 'iPhone 15 Pro Max'`
- `sf lightning preview app -o MyOrg -n MyApp -t android -i 'Pixel 5 API 34'`

@W-15813034@
@W-15813106@
@W-15001240@
@W-15578370@

Screenshot 2024-06-05 at 1 45 53 PM


messages/lightning.preview.app.md
```diff
@@ -39,6 +39,88 @@ For mobile virtual devices, specify the device ID to preview. If omitted, the fi
...
+# certificate.installation.steps.android
+
+1. Drag & drop the file on to your booted emulator.
+2. %s
+3. Browse to the certificate file that you transferred from step 1 (usually under `/sdcard/download`).
+4. Follow the on-screen instructions to install it.
+5. Tap on `User credentials` under `Credential storage` and verify that your certificate is listed there.
+6. Tap on `Trusted credentials` under `Credential storage`. Then tap on `USER` and verify that your certificate is listed there.
+
+# certificate.installation.steps.android.nav-target-api-24-25
```
maliroteh-sf, line 84: we really need to find a better way in the future b/c showing detailed instructions to the user is not sustainable on Android since they seem to keep changing their Settings menu very often and instructions become outdated

src/commands/lightning/preview/app.ts
```diff
@@ -116,22 +193,104 @@ export default class LightningPreviewApp extends SfCommand<void> {
...
+    if (!device) {
+      return Promise.reject(new Error(messages.getMessage('error.device.notfound', [deviceId ?? ''])));
+    }
+
+    // 3. Boot the device if not already booted
+    const resolvedDeviceId = platform === Platform.ios ? (device as IOSSimulatorDevice).udid : device.name;
+    const emulatorPort = await PreviewUtils.bootMobileDevice(platform, resolvedDeviceId, logger);
+
+    // 4. Generate self-signed certificate and wait for user to install it
+    const certFilePath = PreviewUtils.generateSelfSignedCert(platform, '~/Desktop/cert');
```
maliroteh-sf, line 228: Once the WI for creating `lwr.config.json` is done, this will be updated to save the cert in the same location as the config file

messages/lightning.preview.app.md
```diff
@@ -39,6 +39,88 @@ For mobile virtual devices, specify the device ID to preview. If omitted, the fi
...
+
+`%s`
+
+Follow the steps below to install the certificate:
+
+%s
+
+# certificate.installation.steps.ios
+
+1. Drag & drop the file on to your booted simulator.
```
khawkins, line 66:
```suggestion
1. Drag and drop the file on to your booted simulator.
```

messages/lightning.preview.app.md
```diff
@@ -39,6 +39,88 @@ For mobile virtual devices, specify the device ID to preview. If omitted, the fi
...
+3. Tap `Close` and navigate to `Settings > General > VPN & Device Management > localhost`.
+4. Tap `Install` in the title bar, in the warning window, and on the install button.
+5. In the `Profile Installed` view, confirm that the profile shows as `Verified` and Tap `Done`.
+6. Now navigate to `Settings > General > About > Certificate Trust Settings`.
+7. Toggle full trust for `localhost` to enable full trust.
+8. In the warning dialog, tap `Continue`.
+
+# certificate.installation.steps.android
+
+1. Drag & drop the file on to your booted emulator.
```
khawkins, line 77:
```suggestion
1. Drag and drop the file on to your booted emulator.
```

src/shared/previewUtils.ts
```diff
@@ -0,0 +1,430 @@
...
+    deviceId?: string,
+    logger?: Logger
+  ): Promise<IOSSimulatorDevice | AndroidVirtualDevice | undefined> {
+    let device: IOSSimulatorDevice | AndroidVirtualDevice | undefined;
+
+    logger?.debug(`Attempting to get mobile device for platform ${platform}`);
+
+    if (deviceId) {
+      logger?.debug(`Attempting to get device ${deviceId}`);
+      device =
```
khawkins, line 62: It's not my favorite thing to have a single method call that basically controls for multiple behaviors that should probably instead be separated. I think there's an opportunity to split these separate functions out through the use of per-platform implementations of a common interface or interfaces. An example with this use case would be:

```typescript
interface PlatformPreview {
  getMobileDevice(
    deviceId?: string,
    logger?: Logger
  ): Promise<IOSSimulatorDevice | AndroidVirtualDevice | undefined>;
}

export class IOSPlatformPreview implements PlatformPreview {
  getMobileDevice(deviceId?: string, logger?: Logger): Promise<any> {
    // Do iOS stuff here.
  }
}

// app.ts
const platformPreview: PlatformPreview = PreviewUtils.platformPreviewFromPlatform(platform);
const device = await platformPreview.getMobileDevice(deviceId, logger);
// etc.
```

If the conditional logic stays as simple as it is now, it's not a big deal. But it has the potential to become challenging to maintain, if it gets more complex.

src/shared/previewUtils.ts
```diff
@@ -0,0 +1,430 @@
...
+    deviceId: string,
+    logger?: Logger
+  ): Promise<number | undefined> {
+    logger?.debug(`Booting device ${deviceId}`);
+
+    let emulatorPort: number | undefined;
+
+    if (platform === Platform.ios) {
+      await IOSUtils.bootDevice(deviceId, true); // will be no-op if already booted
+      await IOSUtils.launchSimulatorApp();
```
khawkins, line 97: This is basically to give the simulator app focus, once booted, right?

src/shared/previewUtils.ts
maliroteh-sf: Yes I agree... but I think that should be done in `lwc-dev-mobile-core`, where we already have such separation of logic per platform... and I have a `TODO` comment at the top of this file to move pretty much everything from here to `lwc-dev-mobile-core` once we're almost done with coding the features for the beta release.

src/shared/previewUtils.ts
maliroteh-sf: On iOS when you boot a simulator, the process is windowless (the simulator process is running but user won't see any window). Once you launch the simulator app, then it creates a window/view for the process of the simulator that was previously launched.

src/commands/lightning/preview/app.ts
```diff
@@ -116,22 +192,116 @@ export default class LightningPreviewApp extends SfCommand<void> {
...
+    let targetOrg: string | undefined;
     if (idx >= 0 && idx < this.argv.length - 1) {
-      orgOpenCommandArgs.push('--target-org', this.argv[idx + 1]);
+      targetOrg = this.argv[idx + 1];
     }
 
-    await this.config.runCommand('org:open', orgOpenCommandArgs);
+    const launchArguments = PreviewUtils.generateDesktopPreviewLaunchArguments(ldpServerUrl, appId, targetOrg);
+
+    await this.config.runCommand('org:open', launchArguments);
```
khawkins, line 202: I don't know if it's possible to use the new `sf` topic accessors in the command string (i.e. `org open`), but if so that would be preferred. The `<topic>:<sub-topic1>:<sub-topic2>` nomenclature is on the way out.

maliroteh-sf: Unfortunately that doesn't work and instead I see this printed to console:
Screenshot 2024-06-10 at 11 19 51 AM

Which doesn't make sense b/c even the error message prints the exact same command but still doesn't recognize it if I use `org open` instead of `org:open` Maybe that's why in [this thread](https://salesforce-internal.slack.com/archives/C06V045BZD0/p1715186066569749?thread_ts=1715185653.113019&cid=C06V045BZD0) they said to use `org:open`

src/shared/previewUtils.ts
```diff
@@ -0,0 +1,430 @@
...
+    deviceId: string,
+    emulatorPort?: number,
+    logger?: Logger
+  ): Promise<boolean> {
+    logger?.debug(`Checking if ${appConfig.id} is installed on device ${deviceId}`);
+    let result = '';
+    try {
+      if (platform === Platform.ios) {
+        result = CommonUtils.executeCommandSync(`xcrun simctl listapps ${deviceId} | grep "${appConfig.id}"`);
+      } else if (emulatorPort) {
```
khawkins, line 268: Should we warn or error or something if `emulatorPort` _is_ null? I see from [here](https://github.com/salesforcecli/plugin-lightning-dev/pull/50/files#diff-63422716264f9435db5ca6dcc6618d1c753bff38930fe5490849883d4df7b5a1R99) that the emulator port may not be returned if the device is already booted? Is that going to cause issues in downstream behavior like here?

src/shared/previewUtils.ts
```diff
@@ -0,0 +1,430 @@
...
+    let emulatorPort: number | undefined;
+
+    if (platform === Platform.ios) {
+      await IOSUtils.bootDevice(deviceId, true); // will be no-op if already booted
+      await IOSUtils.launchSimulatorApp();
+    } else {
+      emulatorPort = await AndroidUtils.startEmulator(deviceId); // will be no-op if already booted
+    }
+
+    if (emulatorPort) {
```
khawkins, line 102: Can we move the messaging of this conditional into the conditional above? I found myself confused as to where `emulatorPort` would be undefined, only to realize that it's a differentiator of the different platform cases. It'd be more readable just to put these logger messages above.

src/shared/previewUtils.ts
khawkins: Oh, I'm realizing that this is the platform differentiator again. Can we change this conditional instead to something like:

```typescript
if (platform === Platform.ios) {
  result = CommonUtils.executeCommandSync(`xcrun simctl listapps ${deviceId} | grep "${appConfig.id}"`);
} else {
  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
  const resolvedEmulatorPort = emulatorPort!
  result = await AndroidUtils.executeAdbCommand(`shell pm list packages | grep "${appConfig.id}"`, resolvedEmulatorPort);
}
```

This is getting back to my previous comment about separating the behaviors per platform. This is a relatively simple code pattern to follow, and yet the unstated inferences were confusing at first glance.

maliroteh-sf: So I'm using `emulatorPort` as a way to differentiator for the platform cases. This will all read much cleaner once I move the functions in this file to `lwc-dev-mobile-core` where we have separate classes/files per platform.

When targeting iOS `emulatorPort` will be `undefined` but when targeting Android it will have a value (even if booting is no-op b/c it is already booted, we still return a valid port).

src/commands/lightning/preview/app.ts
khawkins: Sounds good. I didn't know if it would work or not, but it sounds like no, and probably doesn't bear further investigation than that at this point. I'm fine with it how it is.

src/shared/previewUtils.ts
khawkins: Got it. I wasn't sure the extent of the bigger refactor you had in mind, with these comments. So I figured I'd point them out. Definitely don't spend a lot of time shuffling here, if it's all meant to move below as it were.

src/commands/lightning/preview/app.ts
```diff
@@ -51,61 +60,128 @@ export default class LightningPreviewApp extends SfCommand<void> {
...
+      }
+
+      installationSteps = messages.getMessage('certificate.installation.steps.android', [
+        messages.getMessage(subStepMessageKey),
+      ]);
     }
+
+    let message = messages.getMessage('certificate.installation.description', [certFilePath, installationSteps]);
+
+    // use chalk to format every substring wrapped in `` so they would stand out when printed on screen
```
khawkins, line 103: Haha nice!

src/commands/lightning/preview/app.ts
```diff
@@ -116,22 +192,116 @@ export default class LightningPreviewApp extends SfCommand<void> {
...
+      }
+
+      // 3. Boot the device if not already booted
+      this.spinner.start(messages.getMessage('spinner.device.boot', [device.toString()]));
+      const resolvedDeviceId = platform === Platform.ios ? (device as IOSSimulatorDevice).udid : device.name;
+      const emulatorPort = await PreviewUtils.bootMobileDevice(platform, resolvedDeviceId, logger);
+      this.spinner.stop();
+
+      // 4. Generate self-signed certificate and wait for user to install it
+      // TODO: update the save location to be the same as server config file path
```
khawkins, line 230: That actually poses a good question. Given that the user needs to interact with the cert (i.e. for manual installation), it rules out putting it in a temp directory. I guess in the absence of the user being able to specify, I'd be inclined (in the final form) to simply save it in the current working directory.

This is one of the many features that will be exponentially nicer in a VSCode UX. 😐

messages/lightning.preview.app.md
```diff
@@ -39,6 +39,106 @@ For mobile virtual devices, specify the device ID to preview. If omitted, the fi
...
+
+`%s`
+
+Follow the steps below to install the certificate:
+
+%s
+
+# certificate.installation.steps.ios
+
+1. Drag and drop the file on to your booted simulator.
```
khawkins, line 84: For both platforms, if it doesn't blow out the scope I wonder if we should consider more concise "inline" instructions, providing links to markdown pages on our repo that describe the installation instructions, complete with images? Maybe that could be a stretch goal for beta, or a GA consideration. I'm afraid anything less is likely going to result in confusion for our users.

src/commands/lightning/preview/app.ts
maliroteh-sf: It may be better to save the cert file in the same location as `lwr.config.json` since we have another work item where we also need to update the config file to include cert & key. Bit I think in most (if not all) cases `process.cwd` would be the same as the location of `lwr.config.json`

messages/lightning.preview.app.md
maliroteh-sf: I wanted to do something like that at first but on Android the steps for navigating through the Settings app is different from one API to another. So I wasn't sure how that would work out if we wanted to use markdown pages containing screenshots that would be different depending on the emulator API level.

But if we go with printing the instructions to console like I'm doing in this PR, at least then I can programmatically  determine what is the API level for the user specified emulator and I can dynamically compose the steps.

But in general I think we should try to see if we can find a solution where the user would not need to install the cert manually (if at all possible)

PR #52, opened 2024-06-08
feat: create web server identity token

sfdctaka:
### What does this PR do?
Created a class that generates 256-bit, CSPRNG-generated identity token associated with the local dev server. After the token is created it is saved in the file, `config.json`, which is a file `sf` CLI saves configuration in by default.

I have purposely omitted for now to call the method `IdentityUtils.updateConfigWithIdentityToken()` from any command until it finds an appropriate place to be called.

~~Created a class that generates 256-bit, CSPRNG-generated identity token associated with the local dev server. After the token is created it is saved in the file, `lwr.config.json`, if it exists.~~

~~I have purposely omitted for now to call the method `IdentityUtils.createIdentityToken()` from any command since there is no `lwr.config.json` created anywhere yet.~~

@W-15818521@


src/shared/identityUtils.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import fs from 'node:fs';
+import { randomBytes } from 'node:crypto';
+import { DevServerUtils } from './devServerUtils.js';
+
+class LwrConfigFile {
+  public identityToken?: string;
+}
+
+const BYTES_TO_GENERATE = 32;
```
maliroteh-sf, line 19: I would remove this and instead have `createIdentityToken(byteSize = 32) { ... }`

src/shared/identityUtils.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export class IdentityUtils {
```
maliroteh-sf, line 21: This is the perfect candidate to be moved to [CryptoUtils.ts or lwc-dev-mobile-core](https://github.com/forcedotcom/lwc-dev-mobile-core/blob/main/src/common/CryptoUtils.ts) later. Please add a `TODO` to this file (similar to what I have [here](https://github.com/salesforcecli/plugin-lightning-dev/blob/main/src/shared/previewUtils.ts#L8-L11)) so that later we refactor it.

src/shared/identityUtils.ts
```diff
@@ -0,0 +1,37 @@
...
+  public static async createIdentityToken(): Promise<void> {
+    const lwrConfigFile = DevServerUtils.getServerConfigFileLocation();
+    if (fs.existsSync(lwrConfigFile)) {
```
maliroteh-sf, line 24: Isn't it necessary to have the identity stored in the config file otherwise preview will not work? If that is the case then the only time this method should resolved would be when it successfully updates the config file. All other cases should throw.

@khawkins is my understanding correct here?

src/shared/identityUtils.ts
```diff
@@ -0,0 +1,37 @@
...
+      const config = DevServerUtils.fetchServerConfigFileContent() as LwrConfigFile;
+      if (config?.identityToken == null) {
```
maliroteh-sf, line 26: Again as stated above, if for some reason we cannot read the content of the config file (i.e `config==null || config==undefined) then that should be an error case. So maybe update the code here like:
```
  public static async createIdentityToken(byteSize = 32): Promise<void> {
    const config = await DevServerUtils.fetchServerConfigFileContent() as LwrConfigFile;
    if (config && !config.identityToken) {
        config.identityToken = randomBytes(byteSize).toString();
        await DevServerUtils.writeServerConfigFileContent(config);
    }
  }
```

This way, if any of the `await ...` statements throws an error (config file not found, unable to read/write config file, etc) then the exception will be bubbled up to the caller (without wrapping it in another Error obj) and the code would be simpler?

test/shared/identityUtils.test.ts
```diff
@@ -0,0 +1,73 @@
...
+import { IdentityUtils } from '../../src/shared/identityUtils.js';
+
+describe('identityUtils', () => {
+  const $$ = new TestContext();
+
+  afterEach(() => {
+    $$.restore();
+  });
+
+  it('createIdentityToken resolves if lwr.config.json is not found', async () => {
```
maliroteh-sf, line 21: Depending on how we decide to update `createIdentityToken` the tests here would need to be updated.

src/shared/identityUtils.ts
maliroteh-sf: Also usually we should be considering bit size not byte size. I know that for our particular case we want 256-bit = 8-byte size but is it always going to be the case that we want to generate as a factor of 8-bits/1-byte ?

maliroteh-sf: I guess since the method is called `createIdentityToken` then it is ok to have it as a factor of 8-bits/1-byte

src/shared/identityUtils.ts
maliroteh-sf: Also maybe we should rename the method to something that is more descriptive of what it is doing? Since the method will generate token and save to file, maybe rename to `updateServerConfigFileWithIdentityToken` ?

src/shared/identityUtils.ts
sfdctaka: I can add a default parameter and set it to 32. Is the reason you want to parameterize is because you want to be able to control the size of token created? If so then I understand. But also then I see it as a future proofing that's not needed at this time when looking at it with the lens of how often do callers of this method be looking at a fairly large token string in 256-bit.

Another question that I have. Are there times when there are more than 8 bits into a byte?

src/shared/identityUtils.ts
sfdctaka: The extension to that argument, I would argue, is that if an identity token already exists, then, do nothing(resolve) as it already satisfies the condition of needing to have an identity token. Maybe the method name of `createIdentityToken()` isn't perfect as it implies that it needs to create a token every time it is getting called.

src/shared/identityUtils.ts
maliroteh-sf: 1. We should try to avoid using global variables (even class level variables) when it may not be necessary. I think in this case, passing as a param with default value helps us remove the global variable and also makes the method generic enough to accept different byte sizes.

2. There will always be 8-bits in a byte. My question around the bits/bytes was more in relations to the fact that this method will always generate a token that is a number of bytes long. Will it ever be the case that we would want to call a method to generate a token that is 134-bits long (which is not divisible by 8)? That was my question.

But then again since the method is called `createIdentityToken` (and uses ` randomBytes` to generate) then it is ok that this method always returns a token with a length that is a factor of 8. So you can ignore the comment regarding bit/byte size.

src/shared/identityUtils.ts
maliroteh-sf: Oh yes I get that part. In fact the suggested method implementation that I've provided [here](https://github.com/salesforcecli/plugin-lightning-dev/pull/52#discussion_r1633931128) accounts for the case where `identityToken` already exists but with less if-statements and logic.

sfdctaka:
@maliroteh-sf For now I went with your idea on handling the error case with config file. All of your comments are addressed. If @khawkins have another idea we can take care of it in this PR or in a later one. As long as you're not blocked.

src/shared/identityUtils.ts
```diff
@@ -0,0 +1,30 @@
...
+
+// **********************************************************************************************
+// * TODO: When we finalize the implementation for the preview commands and things settle down, *
+// *       consider moving this class into CryptoUtils of lwc-dev-mobile-core instead.          *
+// **********************************************************************************************
+export class IdentityUtils {
+  public static async updateServerConfigFileWithIdentityToken(byteSize = 32): Promise<void> {
+    const config = await DevServerUtils.fetchServerConfigFileContent() as LwrConfigFile;
+    if (config && !config.identityToken) {
+      config.identityToken = randomBytes(byteSize).toString();
```
maliroteh-sf, line 26: This didn't occur to me earlier until now. `randomByte` can return bytes that don't represent 'normal' characters so I don't think we can just convert to string and write to file. Based on documentation from [here](https://nodejs.org/api/crypto.html#cryptorandombytessize-callback) we should at least use `.toString('hex')`

Not sure how this will impact the server parsing this value though.

sfdctaka: Aah, great catch. I actually noticed this when I was testing last week and thought I changed it to 'hex'. Let me put it back.

src/shared/devServerUtils.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export class DevServerUtils {
+  public static getServerConfigFileLocation(): string {
+    // TODO: update this after dev server is integrated into this repo
+    return path.join(process.cwd(), 'lwr.config.json');
+  }
+
+  public static fetchServerConfigFileContent(): unknown {
+    const filePath = DevServerUtils.getServerConfigFileLocation();
+    return CommonUtils.loadJsonFromFile(filePath) as unknown;
```
khawkins, line 19: Normally I'd say to create a type around this data instead of using `unknown`. But I think server configuration persistence and access is enough in flux that this won't represent the final state anyway. More to come on that front. Cc: @rax-it

khawkins:
Can we rebase against `main`? It would be good to see these changes in relation to the dev server integration changes that came in from https://github.com/salesforcecli/plugin-lightning-dev/pull/49.

src/shared/identityUtils.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { randomBytes } from 'node:crypto';
+import { DevServerUtils } from './devServerUtils.js';
+
+class LwrConfigFile {
+  public identityToken?: string;
+}
+
+// **********************************************************************************************
+// * TODO: When we finalize the implementation for the preview commands and things settle down, *
```
khawkins, line 19: IMO we shouldn't wait for things to settle down, for this functionality. The heart of this functionality is just a service that creates a CSPRNG-strong token. That core is not going to change, and should be homed in `CryptoUtils`. How we consume it and apply it to the dev server can remain in this package, and we can finalize those details when we know how we're storing the data.

src/shared/identityUtils.ts
khawkins: A few thoughts here:

- In the context of [my previous comment](https://github.com/salesforcecli/plugin-lightning-dev/pull/52/files#r1636897993), the heart of this should move into its own method in `CryptoUtils` in `lwc-dev-mobile-core`.
- This can simply be a function that creates and returns a random token value, based on an input number of bytes. Default could be `32`.
- The return value should be a base64-encoded string, to align with the other areas where we'll use base64 encoding for these values. So i.e. return `randomBytes().toString('base64')`.

maliroteh-sf (approved):

src/shared/devServerUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Copyright (c) 2024, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import path from 'node:path';
+import { CommonUtils } from '@salesforce/lwc-dev-mobile-core';
+
+export class DevServerUtils {
```
sfdctaka, line 11: @maliroteh-sf @khawkins We don't need this class anymore if we are completely switching to leverage `sf config` to stash identity token and any other properties.

src/shared/lightningDevConfig.ts
```diff
@@ -0,0 +1,16 @@
...
+
+import { ConfigFile } from '@salesforce/core';
+
+export const LIGHTNING_DEV_CONFIG = 'lightningDevConfig.json';
+
+export class LightningDevConfig extends ConfigFile {
+  public static getFileName(): string {
+    return LIGHTNING_DEV_CONFIG;
+  }
+}
```
sfdctaka, line 16: I have this class because I couldn't get the default config file to work.

Default config is a file named `config.json` that lives in sfdx project's `.sf` folder. The properties can be read from there, however, not be written. When I try to write I get the error as shown in the screen shot. This error doesn't happen if I specify a different file.
image

sfdctaka: This is what I am doing, @mshanemc

src/shared/identityUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    if (identityToken) {
+      const identityTokenAsString = identityToken as string;
+      return Promise.resolve(identityTokenAsString);
+    }
+
+    return Promise.resolve(undefined);
+  }
+
+  public static async writeIdentityToken(token: string): Promise<void> {
```
sfdctaka, line 42: This method could be `private` but when I do that I won't be able to stub it in the unit test.

src/shared/devServerUtils.ts
maliroteh-sf: Yes we should remove these if not needed anymore

src/shared/identityUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export class IdentityUtils {
+  public static async updateConfigWithIdentityToken(): Promise<void> {
+    const token = await this.getIdentityToken();
+    if (!token) {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment
+      const generatedIdentityToken = CryptoUtils.generateIdentityToken();
+      await this.writeIdentityToken(generatedIdentityToken);
+    }
+    return Promise.resolve();
```
maliroteh-sf, line 24: nit: you don't need `return Promise.resolve();` for a function with return type `Promise<void>` as it is automatically added by the compiler

src/shared/identityUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+    });
+    await config.read();
+    const identityToken = config.get(ConfigVars.LOCAL_WEB_SERVER_IDENTITY_TOKEN);
+
+    if (identityToken) {
+      const identityTokenAsString = identityToken as string;
+      return Promise.resolve(identityTokenAsString);
+    }
+
+    return Promise.resolve(undefined);
```
maliroteh-sf, line 39: Can simply have `return (identityToken as string);` and if `identityToken` is undefined then the casting will also return undefined

src/shared/identityUtils.ts
maliroteh-sf: I think it is ok to leave it `public`.... it is a function in a utility library that could be on use in the future

src/shared/identityUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+      await this.writeIdentityToken(generatedIdentityToken);
+    }
+    return Promise.resolve();
+  }
+
+  public static async getIdentityToken(): Promise<string | undefined> {
+    const config = await LightningDevConfig.create({
+      isGlobal: false,
+    });
+    await config.read();
```
maliroteh-sf, line 31: I would update the method signature to take an optional `config` param so that we don't have to re-read the config if it was already read before. Then you can check to see if `config` param was passed in or not and if not only then proceed with  these lines. This way we can reduce disk access/read

src/shared/identityUtils.ts
maliroteh-sf: I would however update the method signature to take an optional `config` param in addition to the token.

src/shared/identityUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  public static async writeIdentityToken(token: string): Promise<void> {
+    const config = await LightningDevConfig.create({
+      isGlobal: false,
+    });
+    await config.read();
+    config.set(ConfigVars.LOCAL_WEB_SERVER_IDENTITY_TOKEN, token);
+    config.writeSync();
+    return Promise.resolve();
```
maliroteh-sf, line 49: same here, no need for `return Promise.resolve();`

src/configMeta.ts
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright (c) 2024, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import type { ConfigValue, ConfigPropertyMeta } from '@salesforce/core';
+
+export enum ConfigVars {
```
abdulsattar, line 10:
```suggestion
export enum const ConfigVars {
```
https://www.typescriptlang.org/docs/handbook/enums.html#const-enums are cleaner almost always.

src/shared/devServerUtils.ts
sfdctaka: Removed it.

src/shared/identityUtils.ts
sfdctaka: @maliroteh-sf I didn't add optional `config` in "write" because now we need to deal separate "configs" depending on "writing" to or "reading" from `sf`'s configuration. Making `config` optional and delegate that responsibility to the called is certainly an approach that could have been taken. But with given inflexibility with `Config`'s APIs and their shortcomings I think the safest way for now for us is to encapsulate the different  configs(`Config` and `ConfigAggregator`) in respective "write" and "read" methods.

src/shared/identityUtils.ts
```diff
@@ -0,0 +1,37 @@
...
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment
+      const generatedIdentityToken = CryptoUtils.generateIdentityToken();
+      await this.writeIdentityToken(generatedIdentityToken);
+    }
+  }
+
+  public static async getIdentityToken(): Promise<string | undefined> {
+    const config = await ConfigAggregator.create({ customConfigMeta: configMeta });
+    // Need to reload to make sure the values read are decrypted
+    await config.reload();
```
sfdctaka, line 25: @maliroteh-sf @khawkins Need to use `ConfigAggregator` instead of simple `Config` to read the token. The reason is because `Config`'s `get(<token name>, <decrypt flag>)` fails to decrypt the value read from the config file. And the only way is to use `ConfigAggregator` and then to `reload()`.

src/shared/identityUtils.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export class IdentityUtils {
+  public static async updateConfigWithIdentityToken(): Promise<void> {
+    const token = await this.getIdentityToken();
+    if (!token) {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment
+      const generatedIdentityToken = CryptoUtils.generateIdentityToken();
+      await this.writeIdentityToken(generatedIdentityToken);
+    }
+  }
```
abdulsattar, line 20:
```suggestion
  public static async updateConfigWithIdentityToken(): Promise<string> {
    let token = await this.getIdentityToken();
    if (!token) {
      // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment
      const generatedIdentityToken = CryptoUtils.generateIdentityToken();
      token = await this.writeIdentityToken(generatedIdentityToken);
    }
    return token;
  }
```
Could you return the token here? This will be the function that we'll use always to get the identity token. If it doesn't exist, it'll create one and return it.

src/shared/identityUtils.ts
maliroteh-sf: Hmm... that's too bad. Probably need to follow up in [platform-cli-collaboration](https://salesforce.enterprise.slack.com/archives/C06V045BZD0) to see if this is a known limitation, a bug, or are we missing something. But that can be done in the future too.

maliroteh-sf (approved):

src/shared/identityUtils.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import { CryptoUtils } from '@salesforce/lwc-dev-mobile-core';
+import { Config, ConfigAggregator } from '@salesforce/core';
+import configMeta, { ConfigVars } from './../configMeta.js';
+
+export class IdentityUtils {
+  public static async getOrCreateIdentityToken(): Promise<string> {
+    let token = await this.getIdentityToken();
+    if (!token) {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment
```
khawkins, line 16: These disablements don't seem to be necessary anymore.

khawkins (approved):

src/shared/identityUtils.ts
sfdctaka: I pinged one of the engineers earlier and was offered a potential solution but have got to try it yet. Not sure if it will work. I'll make another PR if it does and see if I could address your concern about being able to pass the config.

PR #66, opened 2024-06-20
Pass Identity token to LWC Dev Server

abdulsattar:
### What does this PR do?
To be merged after #52 is merged.

It passes the Identity token to the LWC Dev Server


maliroteh-sf (approved):
Looks like #67 went in before this one, so you have some conflicts to resolved but other than that LGTM

src/lwc-dev-server/index.ts
```diff
@@ -72,6 +74,7 @@ async function createLWCServerConfig(rootDir: string, logger: Logger): Promise<S
     paths: namespacePaths,
     workspace: await LwcDevServerUtils.getLocalDevServerWorkspace(),
     targets: ['LEX'], // should this be something else?
+    identityToken,
```
maliroteh-sf, line 77: in order to be aligned with the rest of the way we pass the params, maybe we should update this to `identityToken: identityToken" instead?

PR #64, opened 2024-06-18
fix: CLI command text review

moorejacqueline:
### What does this PR do?

Suggest revisions to `--help` text for `lightning preview app` and `lightning preview site` to make sure content aligns with CX standards.

### What issues does this PR fix or reference?
[@W-15924073@](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001tShRVYA0/view)

PR #72, opened 2024-06-25
feat: configure LDP server with certificate + some cleanup

maliroteh-sf:
### What does this PR do?

- Add the ability to configure lwc-dev-server server with certificate
- Some refactoring and code cleanup

@W-15906501@


messages/lightning.preview.app.md
```diff
@@ -63,15 +67,17 @@ Preparing to download
...
 
-╔═══════════╗
-║ ATTENTION ║
-╚═══════════╝
+Do you want to skip this step
 
 # certificate.installation.description
 
-If you have not done so already, please install the self-signed certificate on your device before proceeding. The certificate file is located at
+Please install the self-signed certificate on your device before proceeding. The certificate file is located at
```
maliroteh-sf, line 81: @moorejacqueline As I mentioned in a [comment on your PR](https://github.com/salesforcecli/plugin-lightning-dev/pull/64#discussion_r1645232377), I updated the strings to address a UX feedback that we had received. The feedback was to first show a message to the user indicating the need for installing a certificate (one time action) and ask if they've done this already and want to skip or not. If they choose to not skip then show instructions on how to install.

Please review these strings and let me know if anything needs to change.

src/shared/configUtils.ts
```diff
@@ -0,0 +1,77 @@
+/*
```
maliroteh-sf, line 1: Consolidating `identityUtils.ts` and `lwcDevServerUtils.ts` into `configUtils.ts` since the code in both of those files were around reading/writing config values.

messages/shared.utils.md
```diff
@@ -1,19 +1,27 @@
...
 
 The Base64-encoded identity token of the local web server
+
+# config-utils.https-desc
+
+The path to the certificate and private key files to be used by the local dev server for secure connections
+
+# config-utils.https-error-message
+
+Must provide a valid path to the certificate and private key files
```
khawkins, line 27:
```suggestion
You must provide a valid path to the certificate and private key files
```

src/commands/lightning/preview/app.ts
```diff
@@ -171,13 +173,13 @@ export default class LightningPreviewApp extends SfCommand<void> {
...
     } else {
-      await this.mobilePreview(platform, ldpServerUrl, appName, appId, deviceId, logger);
+      await this.mobilePreview(platform, sfdxProjectRootPath, appName, appId, deviceId, logger);
     }
   }
 
-  private async desktopPreview(ldpServerUrl: string, appId?: string, logger?: Logger): Promise<void> {
+  private async desktopPreview(sfdxProjectRootPath: string, appId: string | undefined, logger: Logger): Promise<void> {
     if (!appId) {
       logger?.debug('No Lightning Experience application name provided.... using the default app instead.');
```
khawkins, line 184: Should be able to take the `?` off of `logger` here.

src/commands/lightning/preview/app.ts
```diff
@@ -202,46 +204,68 @@ export default class LightningPreviewApp extends SfCommand<void> {
       targetOrg = this.argv[idx + 1];
     }
 
+    // Start the LWC Dev Server
+    const lwcServer = await startLWCServer(sfdxProjectRootPath, logger);
+    const serverPort = lwcServer.context?.config?.port ?? LOCAL_DEV_SERVER_DEFAULT_PORT;
```
khawkins, line 209: That feels like the long way around to retrieving information that the plugin itself is providing. Should we not read this value from our configuration?

khawkins: But then a lot of the stuff around the setup of the dev server feels disconnected, in general. We've hidden a lot of "public API" details inside of function implementations that initialize not-objects.

It feels to me like:

- The dev server should be an object/type/class: say `LWCDevServer`.
- It should be able to take configuration items that exist in its public API (e.g. ports, project directories, loggers), in its constructor.
- It should use the implementation of its constructor for a lot of the bootstrapping that otherwise happens through code in `createLWCServerConfig()`.

That way you don't end up in situations where you have to query the object you just created, to get the data back the you ostensibly would have otherwise passed into it.

Sorry, probably not actionable in this PR. But that section could use some refactor.

src/configMeta.ts
```diff
@@ -24,6 +26,11 @@ export const enum ConfigVars {
    */
   LOCAL_WEB_SERVER_IDENTITY_TOKEN = 'local-web-server-identity-token',
 
+  /**
+   * The path to certificate and private key files
+   */
+  LOCAL_WEB_SERVER_HTTPS_CERT_AND_KEY_FILES = 'https',
```
khawkins, line 32: Can the value here be something less generic? E.g. `local-web-server-certificate-files`?

src/commands/lightning/preview/app.ts
```diff
@@ -202,46 +204,68 @@ export default class LightningPreviewApp extends SfCommand<void> {
...
-      await LightningPreviewApp.waitForUserToInstallCert(platform, device, certFilePath);
+      // Configure certificates for dev server secure connection
+      let secureConnectionFiles = await ConfigUtils.getSecureConnectionFiles();
+      if (!secureConnectionFiles?.pemCertFilePath || !secureConnectionFiles.pemKeyFilePath) {
+        this.spinner.start(messages.getMessage('spinner.cert.gen'));
+        // Generate self-signed certificate
+        // todo: should the cert files be saved at the root of sfdx project or inside .sf folder of the project?
+        secureConnectionFiles = PreviewUtils.generateSelfSignedCert(platform, sfdxProjectRootPath);
+        // Configure the local dev server with certFilePath & keyFilePath
+        await ConfigUtils.writeSecureConnectionFiles(secureConnectionFiles);
```
khawkins, line 259: @ravijayaramappa How did we decide we were going to handle the interoperability of different user agents needing HTTP vs. HTTPS? That would be a restart of the dev server at this point, yes? I think we need to offset this "write" with a "delete" when HTTPS is _not_ desired. Because the configuration is persisted.

Thinking through that lens, @maliroteh-sf rather than inferring the HTTPS configuration (or lack thereof) based on anticipated use case, we should probably create some helper functionality the decides whether or not we need cert configuration, based on the assessment of (or similar logic to) the WS URI we need.

src/configMeta.ts
maliroteh-sf: Sure

src/commands/lightning/preview/app.ts
maliroteh-sf: Hmmm.... I'm thinking that we need to generate the cert files only once. The cert is valid for ~2 years and can be reused over and over for different cases. For example, if the user wants to preview on iOS and then on Android, they should be able to use the same cert file.... and once they configure their ios/android device they won't need to do that again for another 2 years since we are reusing the same cert file.

On the same note, I think that the cert file should `not` be stored at the root of an SFDX project. This is b/c by doing so we are essentially generating new cert file for different SFDX projects.

 So suppose the user has already generated a cert file for SFDX project1 and installed it on their iOS device. Then they want to preview another SFDX project but on the same iOS device. If we generate a new cert file for SFDX project2, now the user needs to uninstall the previous cert file from the device and install the new one before proceeding.
 
 If we generate the cert files at a global location (say `~/.sf`) then we can reuse the same cert for the next 2 years. We can then follow the logic below:
 
 1. If there are no certificates in `~/.sf` then go ahead and generate them (`localhost.der` + `localhost.pem` + `localhost-key.pem`) regardless of whether we are running for desktop or mobile. These cert files can be used for the next 820 days
 
 2. Generate the LDP server url.
 
 3. If LDP server url starts with `wss://` then we need a secure connection, so use the generated certs in `~/.sf`
 
 4. Refactor the code [here](https://github.com/salesforcecli/plugin-lightning-dev/blob/main/src/lwc-dev-server/index.ts#L82-L102) to something like below, and this way we don't even need to use `configMeta + configUtils` to save the location of the cert files to the config of an SFDX project and it won't be persisting, b/c instead we will create an instance of `ServerConfig` on the fly each time we want to start the dev server:
 ```
 export async function startLWCServer(rootDirOrConfig: string | ServerConfig, logger: Logger): Promise<LWCServer> {
  const finalConfig = 
    (typeof(rootDirOrConfig) === 'string') 
      ? await createLWCServerConfig(rootDirOrConfig, logger)
      : rootDirOrConfig;
  
  logger.trace(`Starting LWC Dev Server with config: ${JSON.stringify(finalConfig)}`);
  let lwcDevServer: LWCServer | null = await startLwcDevServer(finalConfig);
  .
  .
  .
}

khawkins: That's a good point, with regards to having to reconfigure certificate trust all the time for devices on the host machine. For that reason we probably do want to save these values globally.

I'm thinking out loud about a hybrid model of what you suggest, where we both:

- Generate / persist the certificate files once, in a global location.
- _Copy_ these to each _local_ project as necessary, from the global configuration.

The reason I'm thinking of the hybrid potential is that the user still needs to "drag and drop" these certificates to their device(s). In that respect, a) I typically don't like to send users down a "hidden" (`.<dir>`) directory path to search for files, and b) having the files _specifically_ close to their project will make it easier for them to find, because of the association with the project they're working on.

In that scenario we could even persist the (base64-encoded, as necessary) contents of the certificate files themselves within our existing configuration model, just with the `--global` scope for that data, and rehydrate local certificate files from there. Then we don't have to mess around with direct file operations outside of our project directory scope. Anyway, it's a thought I had. I don't know if it's a good one. 🙂

But overall, I think your reasoning on this point is good, and it seems like a sound approach to me.

src/commands/lightning/preview/app.ts
```diff
@@ -202,46 +204,68 @@ export default class LightningPreviewApp extends SfCommand<void> {
...
+
+    logger.debug('Determining Local Dev Server url');
+    const ldpServerUrl = PreviewUtils.generateWebSocketUrlForLocalDevServer(Platform.desktop, serverPort);
+    logger.debug(`Local Dev Server url is ${ldpServerUrl}`);
+
     const launchArguments = PreviewUtils.generateDesktopPreviewLaunchArguments(ldpServerUrl, appId, targetOrg);
 
-    // Start the LWC Dev Server
-    await startLWCServer(process.cwd(), logger ? logger : await Logger.child(this.ctor.name));
+    // todo: For the desktop scenario, should we print a message to the terminal screen to inform the user
```
khawkins, line 217: - We probably should.
- We may want to wait until we have a link to documentation that could further describe it. But if you wanted to do a placeholder in this PR, you could just keep it terse:

> Note: Your desktop browser requires additional configuration to trust the local development server. See the docs for more details.

khawkins (commented):
I think, other than the outstanding certificate discussions, this looks good to me.

messages/lightning.preview.app.md
moorejacqueline: Thanks for tagging me! I'll add a few small suggestions to the new text -- I think I can't do it in this thread.

src/shared/previewUtils.ts
```diff
@@ -30,14 +30,51 @@ import {
...
+
+  /**
+   * Return a port number to be used by the local dev server.
+   *
+   * It starts with the default port (8081)) and checks to see if it is in use or not. If
+   * it is in use then we increment the port number by 2 and check if it is in use or not.
+   * This process is repeated until a port that is not in use is found.
+   *
+   * @returns a port number to be used by the local dev server.
+   */
```
maliroteh-sf, line 53: With this approach, we are able to preview multiple SFDX projects at the same time (if customers ever need to) and have each of them use a different port

messages/lightning.preview.app.md
```diff
@@ -63,15 +67,17 @@ Preparing to download
 
 Downloading
 
-# certificate.attention
+# certificate.installation.notice
+
+A self-signed certificate needs to be installed on your device before you can use your device for previewing. This is a one-time setup only which you can skip if you have installed the certificate already.
```
moorejacqueline, line 76: Could something like this be a little more concise and still accurate?

```suggestion
To use local preview on your device, you have to install a self-signed certificate on it. If you previously set up a certificate for your device, you can skip this step.
```

messages/lightning.preview.app.md
```diff
@@ -63,15 +67,17 @@ Preparing to download
...
 
-╔═══════════╗
-║ ATTENTION ║
-╚═══════════╝
+Do you want to skip this step
 
 # certificate.installation.description
 
-If you have not done so already, please install the self-signed certificate on your device before proceeding. The certificate file is located at
+Please install the self-signed certificate on your device before proceeding. The certificate file is located at
```
moorejacqueline, line 84: I think we usually exclude terms like "please" -- how does this look?

```suggestion 
Before proceeding, install the self-signed certificate on your device. The certificate file is located at
```

src/shared/configUtils.ts
```diff
@@ -0,0 +1,96 @@
...
+    config.set(ConfigVars.LOCAL_WEB_SERVER_IDENTITY_TOKEN, token);
+    await config.write();
+  }
+
+  public static async getCertData(): Promise<SSLCertificateData | undefined> {
+    const config = await this.getGlobalConfig();
+    const data = config.get(ConfigVars.LOCAL_DEV_SERVER_HTTPS_CERT_DATA) as SSLCertificateData;
+    // convert base64 to binary DER data
+    if (data?.derCertificate) {
+      data.derCertificate = Buffer.from(data.derCertificate, 'base64').toString('binary');
```
khawkins, line 67: This is a dicey way to manage binary data, keeping it in a `string`. You can see from Node.js's `Buffer` [docs](https://nodejs.org/docs/latest-v20.x/api/buffer.html#buffers-and-character-encodings), the following relevant verbiage:

> The following legacy character encodings are also supported:
> ...
> `'binary'`: Alias for `'latin1'`. The name of this encoding can be very misleading, as all of the encodings listed here convert between strings and binary data. For converting between strings and `Buffer`s, typically `'utf8'` is the right choice.

The proper way to manage pure binary data in memory, with Node, is in a `Buffer`. I propose that `SSLCertificateData.derCertificate` should be a `Buffer`, not a `string`. It actually simplifies operations on that data:

- No converting the data member into a `Buffer`.
- No messing with encoding.
- No need to further configure file operations, because file reads and writes take a `Buffer` as their outputs and inputs, respectively.

khawkins: Ah, following that back through `lwc-dev-mobile-core`, I see it goes all the way back to node-forge and the fact that their representation of a binary (`Bytes`) type is just an alias to `string`. I don't know why they made that choice, but...¯\\_(ツ)_/¯.

I'd still argue that it's cleaner, more concise, and more representative of the nature of the data, to have it as a `Buffer`. But that would require converting it in `lwc-dev-mobile-core`, via `Buffer.from()` where it's [created](https://github.com/forcedotcom/lwc-dev-mobile-core/blob/d974aefb7c85e3b0687167b5f146b362b62cf202/src/common/CryptoUtils.ts#L124). What are your thoughts?

khawkins: Reading through some of https://github.com/digitalbazaar/forge/issues/146, I'm guessing the design decision was made around flexibility of the code (e.g. use outside of Node.js). But we're standardized on Node.js, so I'd be partial to "converting" to the idioms of our platform.

maliroteh-sf: I see what you mean. As you've discovered already, this is b/c of how node-forge works. Specifically [this line](https://github.com/maliroteh-sf/lwc-dev-mobile-core/blob/main/src/common/CryptoUtils.ts#L124) is where we generate the DER data which node-forge returns back as string (`getBytes()` returns `Bytes` and node-forge has `type Bytes = string;`)

I do have a change coming up in `lwc-dev-mobile-core` and can go ahead and make some changes for this too. So your proposal is to essentially do something along the lines of
```
const derCert = forge.asn1.toDer(forge.pki.certificateToAsn1(cert)).getBytes();
const data = Buffer.from(derCert, 'binary');

return {
    derCertificate: data,
    pemCertificate: pemCert,
    pemPrivateKey: privateKey,
    pemPublicKey: publicKey
};
```
Then here in `configUtils` we would convert the buffer to base64 when saving to global config, and conversely convert back from base64 to buffer when reading from global config?

khawkins: 💯 Yeah that's exactly it. And then working with `Buffer`s on the file operations side is more concise as well.

src/commands/lightning/preview/app.ts
```diff
@@ -170,16 +171,39 @@ export default class LightningPreviewApp extends SfCommand<void> {
       logger.debug(`App Id is ${appId}`);
     }
 
+    logger.debug('Determining the next available port for Local Dev Server');
+    const serverPort = PreviewUtils.getNextAvailablePort();
```
khawkins, line 175: Staging the server port here, effectively takes away the user's ability to configure or keep their own port, [here](https://github.com/salesforcecli/plugin-lightning-dev/pull/72/files#diff-a4a21c3afa081450784063b77a94e1912b0ba8002b853cf9ff77580cf3fa8674R78). We should probably move the existing persisted-config value logic into here, or maybe rename `PreviewUtils.getNextAvailablePort()` to `PreviewUtils.getPort()`, and add the logic into there.

khawkins (commented):
The changes look good to me overall. Just the question of binary cert data (which you have in progress) and the configuration of the port.

src/commands/lightning/preview/app.ts
maliroteh-sf: oh good catch, I'll update

src/shared/previewUtils.ts
```diff
@@ -161,41 +209,47 @@ export class PreviewUtils {
...
-      // all other platforms can work with a base64 PEM text file
-      fileName = path.join(basePath, 'certificate.pem');
-      fs.writeFileSync(fileName, cert.pemCertificate);
+    const targetFile =
+      platform === Platform.ios ? path.join(basePath, 'localhost.der') : path.join(basePath, 'localhost.pem');
+
+    // If we have not previously generated the cert files then go ahead and do so
+    if (!fs.existsSync(targetFile)) {
+      if (platform === Platform.ios) {
+        fs.writeFileSync(targetFile, data.derCertificate, { encoding: 'binary' });
```
khawkins, line 246: So here I'd drop the options specifying encoding, because that's an inherently textual configuration, and doesn't apply to a raw `Buffer`.

khawkins (approved):
Just the minor comment around saving the DER file, but otherwise this looks good to me!

PR #89, opened 2024-07-09
feat: save identity data to sf's config and to the server

sfdctaka:
### What does this PR do?
* Saves identity data structure in `sf`'s config
* Saves identity token on server

@W-15428971@

### DIFF
--- a/messages/shared.utils.md
+++ b/messages/shared.utils.md
@@ -16,7 +16,7 @@ Valid workspace value is "SalesforceCLI" OR "mrt"
 
 # config-utils.token-desc
 
-The Base64-encoded identity token of the local web server
+The identity data is a data structure that ties together a single local web server identity token to multiple orgs.
 
 # config-utils.cert-desc
 
--- a/messages/shared.utils.md
+++ b/messages/shared.utils.md
@@ -14,7 +14,7 @@ The workspace name of the local lwc dev server
 
 Valid workspace value is "SalesforceCLI" OR "mrt"
 
-# config-utils.token-desc
+# config-utils.data-desc
 
 The identity data is a data structure that links the local web server's identity token to the user's configured Salesforce orgs.
 
--- a/messages/shared.utils.md
+++ b/messages/shared.utils.md
@@ -16,7 +16,7 @@ Valid workspace value is "SalesforceCLI" OR "mrt"
 
 # config-utils.token-desc
 
-The identity data is a data structure that ties together a single local web server identity token to multiple orgs.
+The identity data is a data structure that links the local web server's identity token to the user's configured Salesforce orgs.
 
 # config-utils.cert-desc
 
--- a/src/commands/lightning/preview/app.ts
+++ b/src/commands/lightning/preview/app.ts
@@ -160,7 +160,11 @@ export default class LightningPreviewApp extends SfCommand<void> {
 
     logger.debug('Configuring local web server identity');
     const connection = targetOrg.getConnection(undefined);
-    const username = connection.getUsername() as string;
+    const username = connection.getUsername();
+    if (!username) {
+      return Promise.reject(new Error(messages.getMessage('error.username')));
+    }
+
     const token = await ConfigUtils.getOrCreateIdentityToken(username, connection);
 
     let appId: string | undefined;
@@ -238,7 +242,7 @@ export default class LightningPreviewApp extends SfCommand<void> {
       this.log(`\n${messages.getMessage('trust.local.dev.server')}`);
     }
 
-    const launchArguments = PreviewUtils.generateDesktopPreviewLaunchArguments(ldpServerUrl, appId, targetOrg);
+    const launchArguments = PreviewUtils.generateDesktopPreviewLaunchArguments(ldpServerUrl, token, appId, targetOrg);
 
     // Start the LWC Dev Server
     await startLWCServer(logger, sfdxProjectRootPath, token, serverPort);
@@ -330,7 +334,12 @@ export default class LightningPreviewApp extends SfCommand<void> {
 
       // Launch the native app for previewing (launchMobileApp will show its own spinner)
       // eslint-disable-next-line camelcase
-      appConfig.launch_arguments = PreviewUtils.generateMobileAppPreviewLaunchArguments(ldpServerUrl, appName, appId);
+      appConfig.launch_arguments = PreviewUtils.generateMobileAppPreviewLaunchArguments(
+        ldpServerUrl,
+        token,
+        appName,
+        appId
+      );
       await PreviewUtils.launchMobileApp(platform, appConfig, resolvedDeviceId, emulatorPort, bundlePath, logger);
     } finally {
       // stop progress & spinner UX (that may still be running in case of an error)
--- a/src/commands/lightning/preview/app.ts
+++ b/src/commands/lightning/preview/app.ts
@@ -21,6 +21,7 @@ import chalk from 'chalk';
 import { OrgUtils } from '../../../shared/orgUtils.js';
 import { startLWCServer } from '../../../lwc-dev-server/index.js';
 import { PreviewUtils } from '../../../shared/previewUtils.js';
+import { ConfigUtils } from '../../../shared/configUtils.js';
 
 Messages.importMessagesDirectoryFromMetaUrl(import.meta.url);
 const messages = Messages.loadMessages('@salesforce/plugin-lightning-dev', 'lightning.preview.app');
@@ -157,13 +158,18 @@ export default class LightningPreviewApp extends SfCommand<void> {
       return Promise.reject(new Error(messages.getMessage('error.no-project', [(error as Error)?.message ?? ''])));
     }
 
+    logger.debug('Configuring local web server identity');
+    const connection = targetOrg.getConnection(undefined);
+    const username = connection.getUsername() as string;
+    const token = await ConfigUtils.getOrCreateIdentityToken(username, connection);
+
     let appId: string | undefined;
     if (appName) {
       logger.debug(`Determining App Id for ${appName}`);
 
       // The appName is optional but if the user did provide an appName then it must be
       // a valid one.... meaning that it should resolve to a valid appId.
-      appId = await OrgUtils.getAppId(targetOrg.getConnection(undefined), appName);
+      appId = await OrgUtils.getAppId(connection, appName);
       if (!appId) {
         return Promise.reject(new Error(messages.getMessage('error.fetching.app-id', [appName])));
       }
@@ -180,12 +186,13 @@ export default class LightningPreviewApp extends SfCommand<void> {
     logger.debug(`Local Dev Server url is ${ldpServerUrl}`);
 
     if (platform === Platform.desktop) {
-      await this.desktopPreview(sfdxProjectRootPath, serverPort, ldpServerUrl, appId, logger);
+      await this.desktopPreview(sfdxProjectRootPath, serverPort, token, ldpServerUrl, appId, logger);
     } else {
       await this.mobilePreview(
         platform,
         sfdxProjectRootPath,
         serverPort,
+        token,
         ldpServerUrl,
         appName,
         appId,
@@ -198,6 +205,7 @@ export default class LightningPreviewApp extends SfCommand<void> {
   private async desktopPreview(
     sfdxProjectRootPath: string,
     serverPort: number,
+    token: string,
     ldpServerUrl: string,
     appId: string | undefined,
     logger: Logger
@@ -233,7 +241,7 @@ export default class LightningPreviewApp extends SfCommand<void> {
     const launchArguments = PreviewUtils.generateDesktopPreviewLaunchArguments(ldpServerUrl, appId, targetOrg);
 
     // Start the LWC Dev Server
-    await startLWCServer(logger, sfdxProjectRootPath, serverPort);
+    await startLWCServer(logger, sfdxProjectRootPath, token, serverPort);
 
     // Open the browser and navigate to the right page
     await this.config.runCommand('org:open', launchArguments);
@@ -243,6 +251,7 @@ export default class LightningPreviewApp extends SfCommand<void> {
     platform: Platform.ios | Platform.android,
     sfdxProjectRootPath: string,
     serverPort: number,
+    token: string,
     ldpServerUrl: string,
     appName: string | undefined,
     appId: string | undefined,
@@ -316,7 +325,8 @@ export default class LightningPreviewApp extends SfCommand<void> {
       }
 
       // Start the LWC Dev Server
-      await startLWCServer(logger, sfdxProjectRootPath, serverPort, certData);
+
+      await startLWCServer(logger, sfdxProjectRootPath, token, serverPort, certData);
 
       // Launch the native app for previewing (launchMobileApp will show its own spinner)
       // eslint-disable-next-line camelcase
--- a/src/commands/lightning/preview/app.ts
+++ b/src/commands/lightning/preview/app.ts
@@ -206,15 +206,17 @@ export default class LightningPreviewApp extends SfCommand<void> {
     const ldpServerUrl = PreviewUtils.generateWebSocketUrlForLocalDevServer(platform, serverPort, logger);
     logger.debug(`Local Dev Server url is ${ldpServerUrl}`);
 
+    const entityId = await PreviewUtils.getEntityId(username);
+
     if (platform === Platform.desktop) {
-      await this.desktopPreview(sfdxProjectRootPath, serverPort, token, username, ldpServerUrl, appId, logger);
+      await this.desktopPreview(sfdxProjectRootPath, serverPort, token, entityId, ldpServerUrl, appId, logger);
     } else {
       await this.mobilePreview(
         platform,
         sfdxProjectRootPath,
         serverPort,
         token,
-        username,
+        entityId,
         ldpServerUrl,
         appName,
         appId,
@@ -228,7 +230,7 @@ export default class LightningPreviewApp extends SfCommand<void> {
     sfdxProjectRootPath: string,
     serverPort: number,
     token: string,
-    username: string,
+    entityId: string,
     ldpServerUrl: string,
     appId: string | undefined,
     logger: Logger
@@ -261,9 +263,9 @@ export default class LightningPreviewApp extends SfCommand<void> {
       this.log(`\n${messages.getMessage('trust.local.dev.server')}`);
     }
 
-    const launchArguments = await PreviewUtils.generateDesktopPreviewLaunchArguments(
+    const launchArguments = PreviewUtils.generateDesktopPreviewLaunchArguments(
       ldpServerUrl,
-      username,
+      entityId,
       appId,
       targetOrg
     );
@@ -280,7 +282,7 @@ export default class LightningPreviewApp extends SfCommand<void> {
     sfdxProjectRootPath: string,
     serverPort: number,
     token: string,
-    username: string,
+    entityId: string,
     ldpServerUrl: string,
     appName: string | undefined,
     appId: string | undefined,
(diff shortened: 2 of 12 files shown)


src/shared/configUtils.ts
```diff
@@ -35,27 +48,29 @@ export class ConfigUtils {
...
+    if (!identityData) {
+      const token = CryptoUtils.generateIdentityToken();
+      const entityId = await this.saveIdentityTokenToServer(token, connection);
+      identityData = new LocalWebServerIdentityData(token);
+      identityData.usernameToServerEntityIdMap[username] = entityId;
+      await this.writeIdentityData(identityData);
+      return token;
+    } else {
+      let entityId = identityData.usernameToServerEntityIdMap[username];
+      if (!entityId) {
```
sfdctaka, line 62: There could be a case where in the machine's `sf`'s config the identity data was saved. But, somehow the server doesn't have the identity token saved anymore. In that case maybe I don't check for `entityId` but instead always save back the token to the server. What do you think?

test/shared/configUtils.test.ts
```diff
@@ -5,74 +5,89 @@
...
-  it('getOrCreateIdentityToken resolves and writeIdentityToken is called when there is no token', async () => {
-    const fakeIdentityToken = 'fake identity token';
-    $$.SANDBOX.stub(ConfigUtils, 'getIdentityToken').resolves(undefined);
+  it('getOrCreateIdentityToken resolves and writeIdentityData is called when there is no identity data', async () => {
+    const stubConnection = $$.SANDBOX.createStubInstance(Connection);
+    $$.SANDBOX.stub(ConfigUtils, 'getIdentityData').resolves(undefined);
     $$.SANDBOX.stub(CryptoUtils, 'generateIdentityToken').resolves(fakeIdentityToken);
-    const writeIdentityTokenStub = $$.SANDBOX.stub(ConfigUtils, 'writeIdentityToken').resolves();
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    $$.SANDBOX.stub(ConfigUtils as any, 'saveIdentityTokenToServer')
```
sfdctaka, line 51: Had to cast `ConfigUtils` as `any` because `saveIdentityTokenToServer` is a private static method. Without the cast sinon has difficulty stubbing a private method.

src/shared/configUtils.ts
```diff
@@ -5,14 +5,27 @@
...
 import configMeta, { ConfigVars, SerializedSSLCertificateData } from './../configMeta.js';
 
 export const LOCAL_DEV_SERVER_DEFAULT_PORT = 8081;
 export const LOCAL_DEV_SERVER_DEFAULT_WORKSPACE = Workspace.SfCli;
 
+export class LocalWebServerIdentityData {
+  public identityToken: string;
+  public usernameToServerEntityIdMap: Record<string, string> = {};
+
+  public constructor(token: string) {
```
maliroteh-sf, line 24: Nit: let the constructor set the map too
```
export class LocalWebServerIdentityData {
  public identityToken: string;
  public usernameToServerEntityIdMap: Record<string, string>;

  public constructor(token: string, map: Record<string, string> = {}) {
    this.identityToken = token;
    this.usernameToServerEntityIdMap = map
  }
}
```

src/shared/configUtils.ts
maliroteh-sf: Makes sense to me, but I'd defer to Kevin on this.

test/shared/configUtils.test.ts
```diff
@@ -5,74 +5,89 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+/* eslint-disable @typescript-eslint/no-unsafe-call */
```
maliroteh-sf, line 9: Same here... let's limit these to only the lines that need them instead of the entire file

src/shared/configUtils.ts
```diff
@@ -35,27 +48,29 @@ export class ConfigUtils {
...
-    if (!token) {
-      token = CryptoUtils.generateIdentityToken();
-      await this.writeIdentityToken(token);
+  public static async getOrCreateIdentityToken(username: string, connection: Connection): Promise<string> {
+    let identityData = await this.getIdentityData();
+    if (!identityData) {
+      const token = CryptoUtils.generateIdentityToken();
+      const entityId = await this.saveIdentityTokenToServer(token, connection);
+      identityData = new LocalWebServerIdentityData(token);
+      identityData.usernameToServerEntityIdMap[username] = entityId;
```
maliroteh-sf, line 57: If you update the constructor as above, then you can just have `identityData = new LocalWebServerIdentityData(token, { username: entityId });`

messages/shared.utils.md
```diff
@@ -16,7 +16,7 @@ Valid workspace value is "SalesforceCLI" OR "mrt"
 
 # config-utils.token-desc
 
-The Base64-encoded identity token of the local web server
+The identity data is a data structure that ties together a single local web server identity token to multiple orgs.
```
khawkins, line 19:
```suggestion
The identity data is a data structure that links the local web server's identity token to the user's configured Salesforce orgs.
```

src/shared/configUtils.ts
```diff
@@ -5,14 +5,27 @@
  * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
```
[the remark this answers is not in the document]
khawkins: When I remove them there are no warnings. Are these still relevant?

src/shared/configUtils.ts
```diff
@@ -35,27 +48,29 @@ export class ConfigUtils {
...
-
-    return identityToken as string;
   }
 
-  public static async writeIdentityToken(token: string): Promise<void> {
-    const config = await this.getConfig();
-    config.set(ConfigVars.LOCAL_WEB_SERVER_IDENTITY_TOKEN, token);
+  public static async writeIdentityData(identityData: LocalWebServerIdentityData): Promise<void> {
+    const config = await this.getGlobalConfig();
+    config.set(ConfigVars.LOCAL_WEB_SERVER_IDENTITY_DATA, JSON.stringify(identityData));
```
khawkins, line 73: This value can be JSON. We should just set the `identityData` value itself, not a parsed/stringified version of it.

src/shared/configUtils.ts
```diff
@@ -35,27 +48,29 @@ export class ConfigUtils {
...
-    const identityToken = config.getPropertyValue(ConfigVars.LOCAL_WEB_SERVER_IDENTITY_TOKEN);
-
-    return identityToken as string;
   }
 
-  public static async writeIdentityToken(token: string): Promise<void> {
-    const config = await this.getConfig();
-    config.set(ConfigVars.LOCAL_WEB_SERVER_IDENTITY_TOKEN, token);
+  public static async writeIdentityData(identityData: LocalWebServerIdentityData): Promise<void> {
+    const config = await this.getGlobalConfig();
```
khawkins, line 72: The identity data shouldn't be global. It should be project-specific.

maliroteh-sf:
@sfdctaka and @khawkins 

In this PR shouldn't we also update `PreviewUtils.generateDesktopPreviewLaunchArguments()` and `PreviewUtils.generateMobileAppPreviewLaunchArguments()` to generate something for `0.aura.ldpServerId`?

sfdctaka:
> @sfdctaka and @khawkins
> 
> In this PR shouldn't we also update `PreviewUtils.generateDesktopPreviewLaunchArguments()` and `PreviewUtils.generateMobileAppPreviewLaunchArguments()` to generate something for `0.aura.ldpServerId`?

Aah. Thought `desktopPreview(...)` and `mobilePreview(...)` were going to take the token and do that for `0.aura.ldpServerId`. I will add it.

src/shared/configUtils.ts
khawkins: In a world where we don't have the Entity ID for an org stored locally, by definition we no longer have any definitive way to connect back to what may or may not have already existed for this project on the app server. In this case we _have_ to write back to the server, and get a new Entity ID, even if the server has a previous entry for this project. In that case, the previous data becomes orphaned, but otherwise the identity is definitively established with the new entity.

That was an intentional design decision: app server entries needed to potentially be throwaway, because keeping them in sync in all cases would otherwise be too onerous. With this design, you just generate a new association and move on.

src/commands/lightning/preview/app.ts
```diff
@@ -157,13 +158,18 @@ export default class LightningPreviewApp extends SfCommand<void> {
       return Promise.reject(new Error(messages.getMessage('error.no-project', [(error as Error)?.message ?? ''])));
     }
 
+    logger.debug('Configuring local web server identity');
+    const connection = targetOrg.getConnection(undefined);
+    const username = connection.getUsername() as string;
```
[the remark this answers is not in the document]
sfdctaka: The assumption was that the user had already authenticated with the org. If that's not the case, and indeed you're right in `getUsername()` returning undefined, then can we error out of the command? Because without having been already authorized and therefore without user name I cannot make identity data or write the token to server. @maliroteh-sf @khawkins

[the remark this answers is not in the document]
khawkins: Yep, if there's no org authorized, then this command is effectively useless. I don't know if having a `Flags.requiredOrg` argument does that boundary checking for you. But at least as an added check, we can error out if there's no username.

src/shared/configUtils.ts
sfdctaka: When I remove parse/stringify I am getting the following errors. Maybe we do want to keep them parsed/stringified?

```
src/shared/configUtils.ts:66:59 - error TS2345: Argument of type 'LocalWebServerIdentityData' is not assignable to parameter of type 'Optional<AnyJson>'.
  Type 'LocalWebServerIdentityData' is not assignable to type 'JsonMap'.
    Index signature for type 'string' is missing in type 'LocalWebServerIdentityData'.

66     config.set(ConfigVars.LOCAL_WEB_SERVER_IDENTITY_DATA, identityData);
```

```
error TS2352: Conversion of type 'string | number | boolean | JsonMap | JsonArray' to type 'LocalWebServerIdentityData' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type 'AnyJson[]' is missing the following properties from type 'LocalWebServerIdentityData': identityToken, usernameToServerEntityIdMap
```

sfdctaka: I think it might be because I have LocalWebServerIdentityData as class. I'm going to change it to `type`.

sfdctaka: Yup, that was it.

test/shared/configUtils.test.ts
maliroteh-sf: Instead of stubbing a private function, you could stub `sobject` and `insert` on `Connection.prototype`

src/shared/configUtils.ts
sfdctaka: As we spoke on slack this is now a `type` instead of `class`.

src/shared/configUtils.ts
```diff
@@ -35,32 +61,36 @@ export class ConfigUtils {
...
-    config.set(ConfigVars.LOCAL_WEB_SERVER_IDENTITY_TOKEN, token);
+    // TODO: JSON needs to be stringified in order for config.write to encrypt. When config.write()
+    //       can encrypt JSON data to write it into config we shall remove stringify().
+    config.set(ConfigVars.LOCAL_WEB_SERVER_IDENTITY_DATA, JSON.stringify(identityData));
     await config.write();
   }
 
   public static async getCertData(): Promise<SSLCertificateData | undefined> {
-    const config = await this.getGlobalConfig();
+    const config = await this.getConfig();
```
maliroteh-sf, line 93: Nit: I should have done this in my previous PR where I introduced `getGlobalConfig`.... but I think it would be a nice touch to cleanup this file and rename `#config` to `#localConfig` and `getConfig` to `getLocalConfig` to better align with the counterpart `getGlobalConfig`

src/shared/previewUtils.ts
```diff
@@ -495,4 +508,18 @@ export class PreviewUtils {
     logger?.debug(`Extracting archive ${zipFilePath}`);
     await CommonUtils.executeCommandAsync(cmd, logger);
   }
+
+  private static async getEntityId(username: string): Promise<string> {
+    const identityData = await ConfigUtils.getIdentityData();
+    let entityId: string | undefined;
+    if (!identityData) {
+      throw new Error(messages.getMessage('error.identitydata'));
```
maliroteh-sf, line 516: Since this is an `async` function with a `Promise` return type, instead of throwing I would recommend returning `Promise.reject(new Error(...))`

src/shared/configUtils.ts
```diff
@@ -7,12 +7,38 @@
...
-import { Config, ConfigAggregator } from '@salesforce/core';
+import { Config, ConfigAggregator, Connection } from '@salesforce/core';
 import configMeta, { ConfigVars, SerializedSSLCertificateData } from './../configMeta.js';
 
+// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
+export interface IdentityTokenService {
+  saveTokenToServer(token: string): Promise<string>;
+}
+
+export class AppServerIdentityTokenService implements IdentityTokenService {
```
maliroteh-sf, line 18: It seems that `AppServerIdentityTokenService` is only used in this file so it should be made private and not exported.

In fact, I think we should get rid of `IdentityTokenService` interface and instead pass a `Connection` object to `getOrCreateIdentityToken` and `getOrCreateIdentityToken` would then use a private helper function to accomplish its task of saving token to server.

Then in your tests, instead of passing in a `TestIdentityTokenService` you could pass in mock `Connection` and also stub out `connection.sobject` and `sobject.insert`

Or did you do it this way b/c you were hitting test issues?

src/shared/configUtils.ts
```diff
@@ -7,12 +7,38 @@
 
 import { Workspace } from '@lwc/lwc-dev-server';
 import { CryptoUtils, SSLCertificateData } from '@salesforce/lwc-dev-mobile-core';
-import { Config, ConfigAggregator } from '@salesforce/core';
+import { Config, ConfigAggregator, Connection } from '@salesforce/core';
 import configMeta, { ConfigVars, SerializedSSLCertificateData } from './../configMeta.js';
 
+// eslint-disable-next-line @typescript-eslint/consistent-type-definitions
```
khawkins, line 13: Meh, just make it a `type` then. I personally think this rule is a bit too heavy-handed on style choices, but we can use a type interchangeably here, so may as well go with it.

src/shared/configUtils.ts
sfdctaka: Yes, this was done intentionally. Remember the issue 
```
TypeError: Attempted to wrap request which is already wrapped 
```
This was caused by sinon not being able to stub `Connection` which @khawkins helped identify. After consulting with him this was the simple approach that accomplished what we need to do.

sfdctaka: Resolving as we talked about this on slack earlier.

src/shared/configUtils.ts
```diff
@@ -35,32 +44,36 @@ export class ConfigUtils {
...
-    return identityToken as string;
   }
 
-  public static async writeIdentityToken(token: string): Promise<void> {
-    const config = await this.getConfig();
-    config.set(ConfigVars.LOCAL_WEB_SERVER_IDENTITY_TOKEN, token);
+  public static async writeIdentityData(identityData: LocalWebServerIdentityData): Promise<void> {
+    const config = await this.getLocalConfig();
+    // TODO: JSON needs to be stringified in order for config.write to encrypt. When config.write()
+    //       can encrypt JSON data to write it into config we shall remove stringify().
```
maliroteh-sf, line 70: Oh, interesting...

maliroteh-sf (approved):

src/shared/configUtils.ts
```diff
@@ -35,32 +44,36 @@ export class ConfigUtils {
...
+      const entityId = await tokenService.saveTokenToServer(token);
+      identityData = {
+        identityToken: token,
+        usernameToServerEntityIdMap: {},
+      };
+      identityData.usernameToServerEntityIdMap[username] = entityId;
+      await this.writeIdentityData(identityData);
+      return token;
+    } else {
+      const entityId = await tokenService.saveTokenToServer(identityData.identityToken);
```
khawkins, line 60: We should first check to see if the Entity ID for this user already exists in the configuration, before writing the data to the server. We should only update the app server if there's no entity association locally already.

sfdctaka: I might have misunderstood your previous [comment](https://github.com/salesforcecli/plugin-lightning-dev/pull/89#discussion_r1671243350). How will we then deal with a situation when the entity, for what ever reason, got deleted on the server?

src/shared/previewUtils.ts
```diff
@@ -162,26 +162,33 @@ export class PreviewUtils {
...
+   * @param username User that is initiating the local preview
    * @param appId An optional app id for a targeted LEX app
    * @param targetOrg An optional org id
    * @param auraMode An optional Aura Mode (defaults to DEVPREVIEW)
    * @returns Array of arguments to be used by Org:Open command for launching desktop browser
    */
-  public static generateDesktopPreviewLaunchArguments(
+  public static async generateDesktopPreviewLaunchArguments(
     ldpServerUrl: string,
+    username: string,
```
khawkins, line 173: This is sort of a "one of these things is not like the other", in that the rest of the input parameters are more or less directly plugged into the functions of this method, whereas `username` is indirectly used as a lookup.

Should we retrieve the Entity ID beforehand, and pass that value in as the input parameter?

khawkins: Seems like it would be more consistent to look up Entity ID (in some capacity) in `app.ts`, somewhere around where the `ldpServerUrl` is [determined](https://github.com/salesforcecli/plugin-lightning-dev/pull/89/files#diff-4d3aa00584581c9698a9cb2dedc862456f2b15805536ccb7460a8869c80e8839R206).

src/shared/configUtils.ts
khawkins: I'd consider getting deleted on the server as a pretty edgy of edge cases—we provide no interface to do that, and users would have to intentionally, manually initiate such action through their own means. In that case, the server should just throw an error, and users can reset their local project state accordingly.

khawkins: Regarding my previous comment, my thinking was more centered around what happens if configuration is lost _locally_, which is more likely because, you know, file system, projects / files move or get deleted, etc. So I wanted a design that could easily recreate the app server association with the local identity, without having hard-to-diagnose sync issues between the local and app server environments.

test/shared/configUtils.test.ts
```diff
@@ -7,68 +7,94 @@
...
 import { ConfigVars } from '../../src/configMeta.js';
 
 describe('configUtils', () => {
   const $$ = new TestContext();
+  const fakeIdentityToken = 'PFT1vw8v65aXd2b9HFvZ3Zu4OcKZwjI60bq7BEjj5k4=';
+  const username = 'SalesforceDeveloper';
+  const fakeEntityId = 'entityId';
+
+  class TestIdentityTokenService implements IdentityTokenService {
+    // eslint-disable-next-line class-methods-use-this, @typescript-eslint/no-unused-vars
```
khawkins, line 23: Ugh. I was going to code up a nit suggestion for doing away with this warning, only to unearth no less than three other unavoidable warnings as I tried to write a solution. Our guardrails for opinionated TS idioms are...a lot.

khawkins (commented):
I think my only outstanding comment was the `username` lookup, but otherwise this all looks fine to me.

khawkins (approved):

PR #118, opened 2024-07-29
feat: use both http and https ports when configuring local dev server

maliroteh-sf:
### What does this PR do?
Latest version of local dev server allows for having both HTTP and HTTPS connections simultaneously. This PR updates the code to account for this.

@W-16363432@


src/shared/previewUtils.ts
```diff
@@ -58,38 +58,53 @@ export class PreviewUtils {
...
+    let httpsPort = httpPort + 1;
     let done = false;
 
     while (!done) {
-      const cmd =
-        process.platform === 'win32' ? `netstat -an | find "LISTENING" | find ":${port}"` : `lsof -i :${port}`;
+      const httpCmd =
+        process.platform === 'win32' ? `netstat -an | find "LISTENING" | find ":${httpPort}"` : `lsof -i :${httpPort}`;
+
+      const httpsCmd =
```
khawkins, line 78: This command is the same as the one above, with a different argument. Can we pull it out into a method?

src/shared/previewUtils.ts
```diff
@@ -58,38 +58,53 @@ export class PreviewUtils {
...
+        // ignore and continue
+      }
+
+      try {
+        httpsResult = CommonUtils.executeCommandSync(httpsCmd)?.trim();
+      } catch {
+        // ignore and continue
+      }
+
+      if (httpResult || httpsResult) {
```
khawkins, line 98: This logic is basically enforcing that these ports not only need to both be open, but sequential to each other. Wouldn't it make more sense to find the first open port, then find the next available one offset from there?

maliroteh-sf: Oh good point, let me do that

khawkins (approved):

PR #175, opened 2024-09-17
feat: establish valid experience site sessions at startup

nrkruk:
### What does this PR do?
- On local-dev startup, simulate the auth flow for establishing a valid SID token for sites via servlets (`/servlet/networks/switch`). Grab the valid token from those requests and append to any proxied requests back to core for the life of the local development session. Requests in the browser may still show asGuest=true but be proxied to your org with valid authentication credentials by the dev server. 

- Create a new environment variable `SITE_GUEST_ACCESS`. When set to true, this will skip attempting to establish a valid authentication session at startup and allow you to access the site as a guest user. 

**Note: This is more of a temporary solution to support sites auth and may evolve as we update the Beta and work towards GA**  

@W-16647027@

### DIFF
--- a/src/shared/experience/expSite.ts
+++ b/src/shared/experience/expSite.ts
@@ -96,15 +96,6 @@ export class ExperienceSite {
   }
 
   /**
-   * Steps to init:
-   * 1) Verify the site is an actual site within this organization
-   * 2) Verify and output specific message if wrong site type is selected
-   * 2) Verify that site has been published with Local Dev perm enabled (i.e. a static resource exists for this site name)
-   * 3) Establish sid token for proxied requests
-   */
-  // public async init(): Promise<void> {}
-
-  /**
    * Esablish a valid token for this local development session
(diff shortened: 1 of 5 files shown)


src/shared/experience/expSite.ts
```diff
@@ -7,6 +7,7 @@
 import fs from 'node:fs';
 import path from 'node:path';
 import { Org, SfError } from '@salesforce/core';
+import axios from 'axios';
```
kmbauer, line 10: Is there a reason we used axios over native node fetch (just curious).

kmbauer (approved):

src/shared/experience/expSite.ts
nrkruk: not really, it was just a little easier to get working initially. Axios has some nice extra features that help with debugging like interceptors. But yeah I could switch over to using fetch I guess

PR #189, opened 2024-09-30
feat: ensure local dev is enabled for an org

maliroteh-sf:
### What does this PR do?
Update the `lightning:dev:app` command to first check and see whether Local Dev is enabled for an org or not before proceeding with the local preview.

@W-16591156@

### DIFF
--- a/src/commands/lightning/dev/app.ts
+++ b/src/commands/lightning/dev/app.ts
@@ -101,6 +101,11 @@ export default class LightningDevApp extends SfCommand<void> {
       return Promise.reject(new Error(messages.getMessage('error.username')));
     }
 
+    const localDevEnabled = await OrgUtils.isLocalDevEnabled(connection);
+    if (!localDevEnabled) {
+      return Promise.reject(new Error(messages.getMessage('error.localdev.not.enabled')));
+    }
+
     const tokenService = new AppServerIdentityTokenService(connection);
     const token = await ConfigUtils.getOrCreateIdentityToken(username, tokenService);
 
--- a/src/commands/lightning/dev/app.ts
+++ b/src/commands/lightning/dev/app.ts
@@ -24,6 +24,7 @@ import { ConfigUtils, IdentityTokenService } from '../../../shared/configUtils.j
 
 Messages.importMessagesDirectoryFromMetaUrl(import.meta.url);
 const messages = Messages.loadMessages('@salesforce/plugin-lightning-dev', 'lightning.dev.app');
+const sharedMessages = Messages.loadMessages('@salesforce/plugin-lightning-dev', 'shared.utils');
 
 export const iOSSalesforceAppPreviewConfig = {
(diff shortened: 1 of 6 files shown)


test/commands/lightning/dev/app.test.ts
```diff
@@ -226,24 +237,6 @@ describe('lightning dev app', () => {
       await verifyMobileThrowsWhenFailedToGenerateCert(Platform.android);
     });
 
-    /* it('waits for user to manually install the certificate', async () => {
```
maliroteh-sf, line 229: These were left over from another PR. These should have been removed b/c they no longer apply.

src/shared/orgUtils.ts
```diff
@@ -38,4 +38,20 @@ export class OrgUtils {
...
+
+  /**
+   * Checks to see if Local Dev is enabled for the org.
+   *
+   * @param connection the connection to the org
+   * @returns boolean indicating whether Local Dev is enabled for the org.
+   */
+  public static async isLocalDevEnabled(connection: Connection): Promise<boolean> {
+    const metadata = await connection.metadata.read('LightningExperienceSettings', 'enableLightningPreviewPref');
+    // casting to any here b/c LightningExperienceSettings type which is defined in 'jsforce'
```
khawkins, line 50: For what it's worth, this is something that TypeScript is well suited to solving with types. As an example:

```typescript
type LightningPreviewMetadataResponse = {
  enableLightningPreviewPref: boolean | undefined;
};

// ...

const flagValue = (metadata as LightningPreviewMetadataResponse).enableLightningPreviewPref;
const localDevEnabled = flagValue ?? false;
return localDevEnabled;
```

It may need some adjustment based on the actual data you get back, but this is the preferred approach over casting to `any`, which is why there's linting violation to push you in this direction.

src/commands/lightning/dev/app.ts
```diff
@@ -101,6 +101,11 @@ export default class LightningDevApp extends SfCommand<void> {
       return Promise.reject(new Error(messages.getMessage('error.username')));
     }
 
+    const localDevEnabled = await OrgUtils.isLocalDevEnabled(connection);
```
khawkins, line 104: @nrkruk suggested that this should be applied across the plugin, to cover the `site` command as well. Should we add that check into this PR?

src/shared/orgUtils.ts
maliroteh-sf: Oh good point, I'll update the code. For some reason it is actually returned back as a `string` and not a `boolean` but regardless, your proposed approach is better than what I've been doing.

src/commands/lightning/dev/app.ts
maliroteh-sf: Oh I wasn't aware of this. Sure I can add it there too.

khawkins (approved):

PR #195, opened 2024-10-07
feat: prompt user to select device type

maliroteh-sf:
### What does this PR do?
If no value is provided for the flag `--device-type` then interactively prompts the user to select a device type.

@W-16911797@


src/commands/lightning/dev/app.ts
```diff
@@ -39,22 +39,6 @@ export const androidSalesforceAppPreviewConfig = {
...
-  public async saveTokenToServer(token: string): Promise<string> {
-    const sobject = this.connection.sobject('UserLocalWebServerIdentity');
-    const result = await sobject.insert({ LocalWebServerIdentityToken: token });
-    if (result.success) {
-      return result.id;
-    }
-    throw new Error('Could not save the token to the server');
-  }
-}
-
```
maliroteh-sf, line 57: Refactored and moved around (between `configUtils` and `PreviewUtils`)

src/shared/configUtils.ts
```diff
@@ -44,27 +40,16 @@ export class ConfigUtils {
...
-      await this.writeIdentityData(identityData);
-      return token;
-    } else {
-      let entityId = identityData.usernameToServerEntityIdMap[username];
-      if (!entityId) {
-        entityId = await tokenService.saveTokenToServer(identityData.identityToken);
-        identityData.usernameToServerEntityIdMap[username] = entityId;
-        await this.writeIdentityData(identityData);
-      }
-      return identityData.identityToken;
```
maliroteh-sf, line 66: IMO this function didn't belong here. `configUtils` only provides functions to save or fetch values for different configs. But this function was doing extra work (try to fetch, if no there then create a new one, then save it, then return it).

Since this is used during previewing, I refactored it to `previewUtils` instead.

Also simply moved `getIdentityData` up to group it with `writeIdentityData`.

src/shared/prompt.ts
```diff
@@ -1,27 +0,0 @@
-/*
```
maliroteh-sf, line 1: I renamed it to `promptUtils` to match the pattern of the rest of the files in that location but for some reason diff is showing it as deleted instead of renamed.

test/lwc-dev-server/index.test.ts
```diff
@@ -40,10 +40,8 @@ describe('lwc-dev-server', () => {
   });
 
   beforeEach(async () => {
-    $$.SANDBOX.stub(ConfigUtils, 'getOrCreateIdentityToken').resolves('testIdentityToken');
```
maliroteh-sf, line 43: These were not needed and we forgot to delete/clean them up in the past PRs so just cleaning up now.

messages/prompts.md
```diff
@@ -0,0 +1,23 @@
...
+
+Select a site
+
+# site.confirm-update
+
+An updated site bundle is available for "%s". Would you like to download and apply the update?
+
+# device-type.title
+
+Which device type would you like to use for preview
```
maliroteh-sf, line 11: In his Figma, Marcelino had this as "Which platform would you like to preview" but that didn't sound right to me b/c `Dekstop` which is one of the choices is not a platform. Besides this questions is presented to the user when they don't provide a value for the `--device-type` flag (we don't have a `--platform` flag). So for these reasons. I reworded this but am open to changing it if needed.

@jshackell-sfdc what do you think?

src/commands/lightning/dev/app.ts
```diff
@@ -71,20 +55,29 @@ export default class LightningDevApp extends SfCommand<void> {
...
   };
 
+  private static async determinePlatform(flags: { 'device-type'?: Platform }): Promise<Platform> {
+    let platform = flags['device-type'];
+
+    if (!platform) {
+      platform = await PromptUtils.promptUserToSelectPlatform();
+    }
+
+    return platform;
```
sfdctaka, line 72: Can this be used instead?

`return platform ?? await PromptUtils.promptUserToSelectPlatform();`

messages/prompts.md
jshackell-sfdc:
```suggestion
Which device type do you want to use for the preview?
```

jshackell-sfdc: yeah, let's stick with "device type", because that's what the name of the flag is, the help mentions, etc.

I made a slight editing suggestion. Cheers!

messages/prompts.md
```diff
@@ -0,0 +1,23 @@
+# site.select
+
+Select a site
+
+# site.confirm-update
+
+An updated site bundle is available for "%s". Would you like to download and apply the update?
```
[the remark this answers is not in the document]
jshackell-sfdc: just a suggestion!  But I think "do you want to" is more direct than "would you like to", and direct is good, IMHO.

PR #196, opened 2024-10-08
feat: prompt user to select mobile device

maliroteh-sf:
### What does this PR do?
If no value is provided for the flag `--device-id` then interactively prompts the user to select a mobile device.

@W-16911876@

### DIFF
--- a/src/shared/promptUtils.ts
+++ b/src/shared/promptUtils.ts
@@ -6,8 +6,14 @@
  */
 import select from '@inquirer/select';
 import { confirm } from '@inquirer/prompts';
-import { Messages } from '@salesforce/core';
-import { Platform } from '@salesforce/lwc-dev-mobile-core';
+import { Logger, Messages } from '@salesforce/core';
+import {
+  AndroidDeviceManager,
+  AppleDeviceManager,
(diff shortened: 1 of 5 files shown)


src/shared/promptUtils.ts
```diff
@@ -44,4 +50,46 @@ export class PromptUtils {
...
+  public static async promptUserToSelectMobileDevice(
+    platform: Platform.ios | Platform.android,
+    logger?: Logger
+  ): Promise<BaseDevice> {
+    const availableDevices =
+      platform === Platform.ios
+        ? await new AppleDeviceManager(logger).enumerateDevices()
+        : await new AndroidDeviceManager(logger).enumerateDevices();
+
+    if (!availableDevices || availableDevices.length === 0) {
```
sfdctaka, line 63: nit: Could this line be shortened to `availableDevices?.length === 0`?

sfdctaka (approved):
LGTM!

src/shared/promptUtils.ts
sfdctaka: Probably not. Coalesce operator returning `null` can by true to `=== 0`.

PR #198, opened 2024-10-09
feat: prompt user to select lightning experince app

maliroteh-sf:
What does this PR do?
If no value is provided for the flag `--name` then interactively prompts the user to select a lightning experience app.

What issues does this PR fix or reference?
@W-16599797@

### DIFF
--- a/src/shared/orgUtils.ts
+++ b/src/shared/orgUtils.ts
@@ -11,18 +11,26 @@ type LightningPreviewMetadataResponse = {
   enableLightningPreviewPref?: string;
 };
 
+export type AppDefinition = {
+  DeveloperName: string;
+  Label: string;
+  Description: string;
+  DurableId: string;
+};
+
 export class OrgUtils {
   /**
-   * Given an app name, it queries the org to find the matching app id. To do so,
-   * it will first attempt at finding the app with a matching DeveloperName. If
-   * no match is found, it will then attempt at finding the app with a matching
-   * Label. If multiple matches are found, then the first match is returned.
+   * Given an app name, it queries the AppDefinition table in the org to find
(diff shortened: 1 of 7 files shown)


src/shared/orgUtils.ts
```diff
@@ -43,6 +51,39 @@ export class OrgUtils {
...
+      const match = appDefinitions.records.find((definition) => definition.DeveloperName === item.Name);
+      if (match) {
+        results.push({
+          DeveloperName: match.DeveloperName,
+          Label: item.Label,
+          Description: item.Description,
+          DurableId: match.DurableId,
+        });
+      }
+    });
```
maliroteh-sf, line 82: Ideally we would do all of this in a single SQL statement using `JOIN` but as I understand it SF SOQL statements don't support `JOIN` so instead I run 2 query and filter in code (since it won't be the case that an org would have thousands and thousands of apps... but rather a handful of apps)

src/shared/previewUtils.ts
```diff
@@ -104,6 +104,41 @@ export class PreviewUtils {
...
+    appName?: string,
+    logger?: Logger
+  ): Promise<string> {
+    if (appName) {
+      logger?.debug(`Determining App Id for ${appName}`);
+
+      // The appName is optional but if the user did provide an appName then it must be
+      // a valid one.... meaning that it should resolve to a valid appId.
+      const appId = await OrgUtils.getAppDefinitionDurableId(connection, appName);
+      if (!appId) {
```
haifeng-li-at-salesforce, line 128: If appName is not valid, should user be prompted too?

maliroteh-sf: That's not part of the UX design so we error out instead (which is consistent with the behavior of the Beta version).

Issue #211, opened 2024-10-22
Open the desktop preview in another browser


mmablom:
**Is your feature request related to a problem? Please describe.**
I work using 2 browsers; Firefox (system default) for all my normal browsing needs and Chrome for all Salesforce development. Opening metadata in the org through the `org open` command can be done in a different browser based on the `browser` parameter. When using that command I can always open Salesforce in Chrome. Currently this plugin always opens the URL in the default browser, which means I have to copy the URL each time.

**Describe the solution you'd like**
Add the same `browser` parameter from the `org open` command to the `lightning dev app` command.

**Describe alternatives you've considered**
Currently I copy-past the URL, which is fine if a bit tedious.

**Additional context**
...

steals:
A highly anticipated feature. @maliroteh-sf any information if this feature is prioritized for some furure releases?

maliroteh-sf:
@steals currently our priority is to deliver the Single Component Preview feature and take it from Beta to GA. As such other requests such as this one are deprioritized. For the time being, you would need to copy the URL yourself or change your default browser (if that's an option).

PR #217, opened 2024-10-23
feat: add parameter for browser to open preview in

mmablom:
### What does this PR do?

Adds a parameter `browser` which can be used to specify which browser the preview URL opens in when previewing on desktop. Corresponds to and is passed through to the `browser` parameter for the `org:open` command.

Implements [#211](https://github.com/salesforcecli/plugin-lightning-dev/issues/211)

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,8 @@ USAGE
   $ sf lightning dev app -o <value> [--flags-dir <value>] [-n <value>] [-t desktop|ios|android] [-i <value>]
 
 FLAGS
+  -b, --browser=<option>      Browser to open the url in when previewing on desktop.
+                              <options: chrome|edge|firefox>
   -i, --device-id=<value>     ID of the mobile device to display the preview if device type is set to `ios` or
                               `android`. The default value is the ID of the first available mobile device.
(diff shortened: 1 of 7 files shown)


maliroteh-sf:
@mmablom Thank you for your initiative to create this PR for issue [#211](https://github.com/salesforcecli/plugin-lightning-dev/issues/211) which you've raised.

Unfortunately we cannot take this PR as is. New feature requests need to have a work item in our internal system for tracking them and the work item will then be discussed and prioritized by our team (among all the other work that we need to do). We do this to ensure that all aspects of work for a new feature is considered and completed (for example, investigate on how 
 `generateWebSocketUrlForLocalDevServer` should handle the case for different browsers, how should the command handle the new `browser` flag in the interactive mode if a value for this flag is not provided, who will be the person working on updating the documentation, etc.)

As such, I am going to close this PR. However I have created an internal work item in our system which references this PR and issue [#211](https://github.com/salesforcecli/plugin-lightning-dev/issues/211) and we will get to it after the team has discussed it and determine what it's priority should be.

CC @khawkins

PR #232, opened 2024-10-31
fix: download site bundle correctly if site name has special chars

pulkit0555:
### What does this PR do?
Fixes the bug when site name contains special chars and we need to download site bundle
@W-16552899@

### DIFF
--- a/src/shared/experience/expSite.ts
+++ b/src/shared/experience/expSite.ts
@@ -179,11 +179,21 @@ export class ExperienceSite {
 
   public async getRemoteMetadata(): Promise<SiteMetadata | undefined> {
     if (this.metadataCache.remoteMetadata) return this.metadataCache.remoteMetadata;
+    // Check if there are special characters in the site name
+    const siteNameHasSpacesOrSpecialChars = hasSpacesOrSpecialChars(this.siteName);
+    const originalSiteName = this.siteName;
+    if (siteNameHasSpacesOrSpecialChars) {
+      const updatedSiteName = replaceSpacesAndSpecialChars(this.siteName);
+      this.siteName = updatedSiteName;
+    }
     const result = await this.org
       .getConnection()
       .query<{ Name: string; LastModifiedDate: string }>(
         `SELECT Name, LastModifiedDate FROM StaticResource WHERE Name LIKE 'MRT_experience_%_${this.siteName}'`
       );
+    // Changing the site name back to original after fetching data from StaticResource so it may not break the flow elsewhere
+    this.siteName = originalSiteName;
+
     if (result.records.length === 0) {
       return undefined;
     }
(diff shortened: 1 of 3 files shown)


src/shared/experience/expSite.ts
```diff
@@ -179,11 +179,21 @@ export class ExperienceSite {
...
+      const updatedSiteName = replaceSpacesAndSpecialChars(this.siteName);
+      this.siteName = updatedSiteName;
+    }
     const result = await this.org
       .getConnection()
       .query<{ Name: string; LastModifiedDate: string }>(
         `SELECT Name, LastModifiedDate FROM StaticResource WHERE Name LIKE 'MRT_experience_%_${this.siteName}'`
       );
+    // Changing the site name back to original after fetching data from StaticResource so it may not break the flow elsewhere
+    // this.siteName = originalSiteName;
```
pulkit0555, line 195: @nrkruk If we uncomment this, it breaks the flow when we try to bring the site up in some cases where site name have special chars other than "-"

src/shared/experience/expSite.ts
```diff
@@ -356,3 +366,27 @@ function getSiteNameFromStaticResource(staticResourceName: string): string {
   }
   return parts.slice(4).join(' ');
 }
+
+export function replaceSpacesAndSpecialChars(value: string): string {
+  // Replace any special characters with underscore
+  value = value.replace(/[^a-zA-Z0-9]/g, '_');
```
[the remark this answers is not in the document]
nrkruk: Is that true of how the site names are saved in the static resources?

src/shared/experience/expSite.ts
nrkruk: Not sure what this means. Does this impact functionality in some way?

nrkruk:
We will also need to backport this to the 252-patch branch

src/shared/experience/expSite.ts
pulkit0555: @nrkruk  I was resetting the site name back to original after executing the query but when it tries to bring up the site it fails. So, basically we will get that site bundle but it won't bring up the site.

Also, should we do the site name updation inside the constructor itself? It's already replacing the space with underscore character.

```
  public constructor(org: Org, siteName: string) {
    this.org = org;
    this.siteDisplayName = siteName.trim();
    this.siteName = this.siteDisplayName.replace(' ', '_');
  }
```

src/shared/experience/expSite.ts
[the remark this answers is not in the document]
pulkit0555: I have tested with below examples of site names, works fine.

- space separated site
- site@with#special-chars
- site-with-hyphen
- byoNoBasePath
- byo enhanced

pulkit0555:
Closing this forked repo PR. Will create a new one.

PR #235, opened 2024-11-01
fix: bump target API version to 63.0 to match main codeline

abdulsattar:
### What does this PR do?
Bumps target API version to 63.0 to track main
@W-17117927@

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "typescript": "^5.5.4"
   },
   "engines": {
-    "node": ">=18.0.0 <22"
+    "node": ">=18.0.0 <23"
   },
   "files": [
     "/lib",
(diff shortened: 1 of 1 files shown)


maliroteh-sf (approved):

package.json
```diff
@@ -34,7 +34,7 @@
     "typescript": "^5.5.4"
   },
   "engines": {
-    "node": ">=18.0.0 <22"
+    "node": ">=18.0.0 <23"
```
maliroteh-sf, line 37: Should this now be `>=20 <23` since 18 is no longer supported?

abdulsattar: I based this on the fact that the cli still supports 18: https://github.com/salesforcecli/cli/blob/8ead09d20a321cd847753e1f27a2ef075c9a5ce9/package.json#L13 and that [Node@18 is in maintenance](https://nodejs.org/en/about/previous-releases)

maliroteh-sf (approved):

PR #262, opened 2024-11-20
fix: download site bundle correctly if site name has special chars

pulkit0555:
### What does this PR do?
Fixes the bug when site name contains special chars and we need to download site bundle

@W-16552899@


src/shared/experience/expSite.ts
```diff
@@ -179,6 +179,13 @@ export class ExperienceSite {
 
   public async getRemoteMetadata(): Promise<SiteMetadata | undefined> {
     if (this.metadataCache.remoteMetadata) return this.metadataCache.remoteMetadata;
+    // Check if there are special characters in the site name
+    const isInvalidSiteName = hasSpacesOrSpecialChars(this.siteName);
+    // const originalSiteName = this.siteName;
+    if (isInvalidSiteName) {
+      const updatedSiteName = replaceSpacesAndSpecialChars(this.siteName);
+      this.siteName = updatedSiteName;
+    }
```
nrkruk, line 188: @pulkit0555 Is there a reason why we can't do this in the constructor?

nrkruk (approved):

PR #275, opened 2024-11-26
Update lwr and @lwrjs/api to latest

nrkruk:
This PR updates both the lwr and @lwrjs/api packages to the newly released version latest.


nrkruk:
testing

PR #390, opened 2025-05-01
@W-18182827 fix: added changes for preview without specifying component name

pulkit0555:
### What does this PR do?
[CLI] Support launching without specifying a component
@W-18182827@

### DIFF
--- a/src/commands/lightning/dev/component.ts
+++ b/src/commands/lightning/dev/component.ts
@@ -88,7 +88,7 @@ export default class LightningDevComponent extends SfCommand<void> {
       rootDir,
       mode: 'dev',
       port,
-      name: name ? `c/${name}` : 'c/',
+      name: `c/${name ?? ''}`,
       namespacePaths,
     });
   }
--- a/src/commands/lightning/dev/component.ts
+++ b/src/commands/lightning/dev/component.ts
@@ -11,7 +11,6 @@ import { SfCommand, Flags } from '@salesforce/sf-plugins-core';
 import { Messages, SfProject } from '@salesforce/core';
 import { cmpDev } from '@lwrjs/api';
 import { ComponentUtils } from '../../../shared/componentUtils.js';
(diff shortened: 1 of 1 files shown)


src/commands/lightning/dev/component.ts
```diff
@@ -79,12 +78,6 @@ export default class LightningDevComponent extends SfCommand<void> {
       }
 
       name = match.name;
-    } else {
```
maliroteh-sf, line 82: This should not be removed.... In the previous release we added the ability to use prompts and ask the user for values to different option flags. This applied to `app`, `site`, and `component` commands and was part of the user experience.

Basically the user should be able to just run `sf lightning dev` and interactively be asked a series of questions to determine what the user wants to do. Of course the user has the option of running `sf lightning dev blah blah` and provide the flag values but if they choose not to provide anything and just run `sf lightning dev` then the interactive experience should take over.

maliroteh-sf, line 82: This should not be removed.... In the previous release we added the ability to use prompts and ask the user for values to different option flags. This applied to `app`, `site`, and `component` commands and was part of the user experience.

Basically the user should be able to just run `sf lightning dev` and interactively be asked a series of questions to determine what the user wants to do. Of course the user has the option of running `sf lightning dev blah blah` and provide the flag values but if they choose not to provide anything and just run `sf lightning dev` then the interactive experience should take over.

maliroteh-sf (changes requested):

src/commands/lightning/dev/component.ts
maliroteh-sf: Are we saying that we will continue to have the interactive behavior for all other commands/parameters and only for this one we decide not to ask for its value b/c the user can instead just pick it from the menu after single component preview interface loads up?

nrkruk:
No longer needed

PR #401, opened 2025-05-15
fix: update version recommendation logic

maliroteh-sf:
### What does this PR do?

Fix a bug in version recommendation logic

@W-17672756@

### DIFF
--- a/package.json
+++ b/package.json
@@ -219,11 +219,13 @@
     }
   },
   "apiVersionMetadata": {
-    "comment": "Refer to ApiVersionMetadata in orgUtils.ts for details",
+    "comment1": "Refer to ApiVersionMetadata in orgUtils.ts for more details.",
+    "comment2": "target section defines the dev server version (matchingDevServerVersion) and the API version that it can support (versionNumber).",
     "target": {
       "versionNumber": "64.0",
       "matchingDevServerVersion": "~11.5.0"
     },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -219,11 +219,13 @@
     }
   },
   "apiVersionMetadata": {
-    "comment": "Refer to ApiVersionMetadata in orgUtils.ts for details",
+    "comment1": "Refer to ApiVersionMetadata in orgUtils.ts for more details.",
```
khawkins, line 222: Nit: this is just organizational semantics to a facility that has no formal support. That said, it might be nice to have a more descriptive structure for these comments.
```
    "x-apiVersionMetadata-comments": [
        "Refer to ApiVersionMetadata in orgUtils.ts for more details.",
        "target section defines the dev server version (matchingDevServerVersion) and the API version that it can support (versionNumber)."
    ],
```

khawkins (approved):
This otherwise seems good to me.

package.json
khawkins: This can really be anything that JSON can support (and that doesn't interfere with the data structure otherwise). `x-*` fields are a typical way to express project-specific information in "headers", so I threw that idea out there.

maliroteh-sf: Oh I like this... will change.

PR #447, opened 2025-07-08
feat: @W-18289065@ org-based component preview

bpbuch:
### What does this PR do?
Moving component preview to an org-based solution using the LWR preview application, and removes the local-only component development flow.

With this transition, the usage of `@lwrjs` for the local dev server is replaced with the `lwc-dev-server`.
[@W-18289065@](https://gus.lightning.force.com/a07EE00002CszEfYAJ)

### DIFF
--- a/src/commands/lightning/dev/component.ts
+++ b/src/commands/lightning/dev/component.ts
@@ -6,15 +6,17 @@
  */
 
 import path from 'node:path';
-import url from 'node:url';
 import { SfCommand, Flags } from '@salesforce/sf-plugins-core';
-import { Messages, SfProject } from '@salesforce/core';
-import { cmpDev } from '@lwrjs/api';
+import { Messages, SfProject, Logger } from '@salesforce/core';
+import { Platform } from '@salesforce/lwc-dev-mobile-core';
 import { ComponentUtils } from '../../../shared/componentUtils.js';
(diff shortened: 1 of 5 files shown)


src/commands/lightning/dev/component.ts
```diff
@@ -63,48 +87,49 @@ export default class LightningDevComponent extends SfCommand<void> {
...
+    const targetOrgArg = PreviewUtils.getTargetOrgFromArguments(this.argv);
+    const launchArguments = PreviewUtils.generateComponentPreviewLaunchArguments(
+      ldpServerUrl,
+      ldpServerId,
+      componentName,
+      targetOrgArg
+    );
+
+    // Open the browser and navigate to the right page
+    await this.config.runCommand('org:open', launchArguments);
```
nrkruk, line 133: @bpbuch I need to be able to get the proper URL and iframe it inside of the VSCode webview. Right now it works because I know the URL is localhost, but with this change how will I be able to get the right URL and set that into the iframe?

nrkruk: Also we still want a way to not automatically launch the browser when running the command in VSCode / Code Builder (i.e. we only want to open the preview panel)

nrkruk (approved):

PR #448, opened 2025-07-09
fix: bump LWC to 13.0.29 @W-18283826

abdulsattar:
### What does this PR do?
Bumps LWC to 13.0.29 (to add `salesforce-experience.com` as a trusted host).
Also fixes checking version logic to use semver rather than comparing strings.
@W-18283826@

### DIFF
--- a/.husky/check-versions.js
+++ b/.husky/check-versions.js
@@ -1,23 +1,24 @@
 import fs from 'node:fs';
+import semver from 'semver';
 
 // Read package.json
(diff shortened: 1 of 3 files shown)


ravijayaramappa (approved):

.husky/check-versions.js
```diff
@@ -1,23 +1,24 @@
 import fs from 'node:fs';
```
ravijayaramappa, line 1: Interesting that husky is not even declared as a devDependency of this project.

PR #449, opened 2025-07-10
feat: launch site preview @W-18283821@

abdulsattar:
### What does this PR do?
Launches site in preview URL
@W-18283821@

### DIFF
--- a/src/commands/lightning/dev/site.ts
+++ b/src/commands/lightning/dev/site.ts
@@ -6,11 +6,15 @@
  */
 import fs from 'node:fs';
 import { SfCommand, Flags } from '@salesforce/sf-plugins-core';
-import { Messages } from '@salesforce/core';
+import { Logger, Messages, SfProject } from '@salesforce/core';
+import { Platform } from '@salesforce/lwc-dev-mobile-core';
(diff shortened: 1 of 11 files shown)


src/commands/lightning/dev/site.ts
```diff
@@ -63,6 +72,46 @@ export default class LightningDevSite extends SfCommand<void> {
...
+
+        const logger = await Logger.child(this.ctor.name);
+        await startLWCServer(logger, sfdxProjectRootPath, ldpServerToken, Platform.desktop, serverPorts);
+        const url = new URL(previewUrl);
+        url.searchParams.set('aura.lwcDevServerUrl', ldpServerUrl);
+        url.searchParams.set('aura.lwcDevServerId', ldpServerId);
+        url.searchParams.set('lwc.mode', 'dev');
+        await open(url.toString());
+        return;
+      }
```
nrkruk, line 113: nit: could we separate this out in into two different function calls? Run method then just calls previewHMR() / previewSSR() based on the cli arg being there?

abdulsattar: Done

nrkruk (approved):

PR #676, opened 2026-07-22
fix: repair post-release dev-preview NUTs (W-23400737)

jhefferman-sfdc:
## What & Why

The dev-preview e2e NUTs added in #626 (`test/commands/lightning/dev/component-preview/*.nut.ts`) pass in this repo's own CI but **fail in the post-CLI-release integration pipeline** (which runs them against a released, bundled CLI). This PR fixes them. Tracked by [@W-23400737@](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002eRzdeYAC/view).

Three layered failures were diagnosed, each masking the next:

### 1. API-version drift (the real, deterministic bug)
Scratch orgs are provisioned on the platform's current default API version (e.g. `68.0`), but the plugin only bundles LWC dev-server dists for the versions in `package.json` `apiVersionMetadata` (`66.0`/`67.0`). `dependencyLoader.ts` does an exact-match check and throws, so the dev server exits **before** printing a preview URL, so every browser test hits the 30s `getPreviewURL` timeout.

**Fix:** added `getMaxSupportedApiVersion()` (reads `package.json apiVersionMetadata`) and spawn the dev server with `--api-version <max>`. This decouples the e2e tests from platform version drift and auto-tracks new versions as they're added to `package.json`. (Adding real `68.0` support to the shipping plugin is a separate product task.)

### 2. Sinon cwd cascade (masking bug)
`TestSession`'s constructor stubs `process.cwd` **before** creating scratch orgs. If org creation throws, `create()` rejects without returning, the stub leaks, and every subsequent file dies with a misleading `Attempted to wrap cwd which is already wrapped` — hiding the true error.

**Fix:** restore the leaked stub on setup failure before rethrowing, so the real failure surfaces in the file that actually failed. Also added `retries` (default 3, overridable via `TESTKIT_SETUP_RETRIES`) for intermittent scratch-org signup (`RemoteOrgSignupFailed` / C-9999).

### 3. Secret redaction, frontdoor 401
`sf org display user --json` now **redacts** `accessToken` by default (returns a `[REDACTED] ...` placeholder), so frontdoor SSO fails with HTTP 401 and the component never renders.

**Fix:** read the token directly from the org connection via `@salesforce/core` (`Org.create().getConnection().accessToken`) instead of shelling out.

## Testing

Full NUT suite verified passing locally against a healthy DevHub, and green in CI on both `ubuntu-latest` and `windows-latest`:

```
lightning preview menu               ✔ ✔
lightning preview component error    ✔ ✔ ✔
lightning preview hot module reload  ✔ ✔ ✔
lightning preview component prompts  ✔ ✔ ✔ ✔

12 passing (2m)
```

Scope: test-helpers only (`test/commands/lightning/dev/helpers/`) — no changes to shipping plugin code.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test/commands/lightning/dev/helpers/browserUtils.ts
+++ b/test/commands/lightning/dev/helpers/browserUtils.ts
@@ -15,35 +15,29 @@
  */
 
 import { chromium, type Browser, type Page } from 'playwright';
-import { execCmd, TestSession } from '@salesforce/cli-plugins-testkit';
-
-type OrgDisplayUserResult = { accessToken?: string };
+import { TestSession } from '@salesforce/cli-plugins-testkit';
+import { Org } from '@salesforce/core';
 
 /**
  * Returns the access token for frontdoor sid authentication. Required for
- * Playwright testing. Uses testkit execCmd so the correct CLI executable is
- * used on all platforms (e.g. sf on Unix, sf.cmd on Windows).
+ * Playwright testing. Reads the token directly from the org connection via the
+ * core Org API rather than shelling out to `sf org display user --json`, which
+ * now redacts secrets by default (returning a "[REDACTED] ..." placeholder
+ * instead of the real token, which would make frontdoor auth fail with a 401).
  *
  * @param session - TestSession with a default scratch org.
  * @returns The session ID string.
  */
-export function getAccessToken(session: TestSession): string {
+export async function getAccessToken(session: TestSession): Promise<string> {
   const scratchOrg = session.orgs.get('default');
   const username = scratchOrg?.username ?? '';
-  const projectDir = session.project?.dir ?? '';
-  if (!username || !projectDir) {
-    throw new Error('Session has no default scratch org username or project dir');
+  if (!username) {
+    throw new Error('Session has no default scratch org username');
   }
-  const result = execCmd<OrgDisplayUserResult>(`org display user -o ${username} --json`, {
-    cwd: projectDir,
-    cli: 'sf',
-    ensureExitCode: 0,
-  });
-  const accessToken = result.jsonOutput?.result?.accessToken ?? '';
+  const org = await Org.create({ aliasOrUsername: username });
+  const accessToken = org.getConnection().accessToken ?? '';
   if (!accessToken) {
-    throw new Error(
-      `sf org display user result missing accessToken: ${result.shellOutput.stdout} ${result.shellOutput.stderr ?? ''}`,
-    );
+    throw new Error(`Could not resolve an access token for org '${username}'`);
   }
   return accessToken;
 }
@@ -72,7 +66,7 @@ async function establishSessionViaFrontDoor(page: Page, previewOrigin: string, a
  * @returns Promise resolving to the Playwright browser and page; caller must close them when done.
  */
 export async function getPreview(previewUrl: string, session: TestSession): Promise<{ browser: Browser; page: Page }> {
-  const accessToken = getAccessToken(session);
+  const accessToken = await getAccessToken(session);
   const previewOrigin = new URL(previewUrl).origin;
   const headed = process.env.HEADED === 'true' || process.env.HEADED === '1';
   const browser = await chromium.launch({ headless: !headed });
--- a/test/commands/lightning/dev/helpers/devServerUtils.ts
+++ b/test/commands/lightning/dev/helpers/devServerUtils.ts
@@ -17,6 +17,7 @@ import { spawn, ChildProcessByStdio, ChildProcessWithoutNullStreams } from 'node
 import { Readable } from 'node:stream';
 import type { Writable } from 'node:stream';
 import { type ChildProcess } from 'node:child_process';
+import fs from 'node:fs';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
 import { TestSession } from '@salesforce/cli-plugins-testkit';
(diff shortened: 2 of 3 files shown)


test/commands/lightning/dev/helpers/devServerUtils.ts
```diff
@@ -28,6 +29,26 @@ const MAX_WAIT_MS = 30_000;
...
+ * and auto-tracks new versions as they are added to package.json.
+ *
+ * @returns The max supported API version string (e.g. '67.0').
+ */
+export function getMaxSupportedApiVersion(): string {
+  const pkgPath = path.join(PLUGIN_ROOT_PATH, 'package.json');
+  const pkg = JSON.parse(fs.readFileSync(pkgPath, 'utf8')) as { apiVersionMetadata?: Record<string, unknown> };
+  const versions = Object.keys(pkg.apiVersionMetadata ?? {});
+  if (versions.length === 0) {
+    throw new Error(`No apiVersionMetadata found in ${pkgPath}`);
```
rax-it, line 47: nit: should we try a default version?

rax-it (approved):

test/commands/lightning/dev/helpers/devServerUtils.ts
jhefferman-sfdc: Intentional fail-fast here rather than a default. `apiVersionMetadata` is a hard requirement for the plugin — the shipping `dependencyLoader.ts` reads `packageJson.apiVersionMetadata` directly and would break too if it were ever empty, so an empty map means the plugin is misconfigured, not in a transient state a test should paper over.

There's also no existing "default API version" constant anywhere in the source to fall back to, so a default would have to be a hardcoded magic number (e.g. `'67.0'`). That would silently rot as versions move, and could pin the tests to a version with no matching `@lwc/sfdx-local-dev-dist-<v>` dist — producing a more confusing downstream import failure than the clear "No apiVersionMetadata found" we throw at the source.

Happy to revisit if we ever introduce a canonical default version the plugin resolves against.

test/commands/lightning/dev/helpers/browserUtils.ts
```diff
@@ -15,35 +15,29 @@
...
+ * Playwright testing. Reads the token directly from the org connection via the
+ * core Org API rather than shelling out to `sf org display user --json`, which
+ * now redacts secrets by default (returning a "[REDACTED] ..." placeholder
+ * instead of the real token, which would make frontdoor auth fail with a 401).
  *
  * @param session - TestSession with a default scratch org.
  * @returns The session ID string.
  */
-export function getAccessToken(session: TestSession): string {
+export async function getAccessToken(session: TestSession): Promise<string> {
```
wjhsf, line 31:
```suggestion
async function getAccessToken(session: TestSession): Promise<string> {
```
not used outside this file, doesn't need to be exported

test/commands/lightning/dev/helpers/sessionUtils.ts
```diff
@@ -22,23 +22,54 @@ let cachedSession: TestSession;
 
 const PROJECT_PATH = path.resolve(PLUGIN_ROOT_PATH, 'test/projects/component-preview-project');
 
+// Number of times TestSession will retry scratch org creation before failing. Scratch org
+// signup is intermittently flaky (RemoteOrgSignupFailed / C-9999); retrying avoids spurious
+// failures in the post-release pipeline. Overridable via TESTKIT_SETUP_RETRIES.
+const SETUP_RETRIES = Number.parseInt(process.env.TESTKIT_SETUP_RETRIES ?? '', 10) || 3;
```
wjhsf, line 28: Curious use of `Number.parseInt` instead of global `parseInt`.

wjhsf (commented):
What happens when the platform gets a new default API version, but this package doesn't get updated?

test/commands/lightning/dev/helpers/browserUtils.ts
jhefferman-sfdc: Good catch — addressed in follow-up #677. `getAccessToken` is now non-exported (it's only used by `getPreview` in this file).

test/commands/lightning/dev/helpers/sessionUtils.ts
jhefferman-sfdc: Switched to the global `parseInt` in #677 for consistency — the rest of the repo (`src/configMeta.ts`, `src/shared/previewUtils.ts`, and `devServerUtils.ts` right next door) uses the global form, so `Number.parseInt` was the odd one out here. No behavioral difference either way.

jhefferman-sfdc:
> @wjhsf: What happens when the platform gets a new default API version, but this package doesn't get updated?

That's precisely the drift this change is built to absorb, so the answer is "the e2e tests keep passing." Before this fix they broke on exactly that scenario.

The mechanism: `startLightningDevServer` no longer lets the org's (newer) default API version reach the dev server. It pins the preview to `getMaxSupportedApiVersion()` — the highest version in the plugin's own `apiVersionMetadata`, i.e. the newest version this package actually ships a `@lwc/sfdx-local-dev-dist-<v>` for. So:

- **Platform bumps its default (e.g. 68.0), plugin not yet updated** → tests still run against the plugin's max supported version (e.g. 67.0), which has a matching dist, and stay green. No spurious post-release failures.
- **A human adds 68.0 to the plugin** (new dist dep + `apiVersionMetadata` entry) → `getMaxSupportedApiVersion()` automatically returns `68.0` and the tests track forward with zero test changes.

The one thing this deliberately does *not* do is give the plugin real support for a newer API version than it ships — that's a separate product task (adding the dist + metadata + types), not something a test should silently paper over. The tests validate what the plugin actually supports today, decoupled from platform drift. Full root-cause writeup is in the #676 description.

